PR #177, opened 2024-12-18
feat(autoware_lanelet_map_validator): add dangling reference checker to non existing intersection_area

soblin:
## Description

This PR aims to add checker if any intersection lanelets (namely with `turn_direction` tag) has valid VALUE as the `intersection_area` key. If there are no intersection_area with corresponding VALUE id, it is **dangling reference**.

## How was this PR tested?

Added a test map "intersection_area_with_dangling_reference.osm" with:
- valid reference to existing intersection_area 10803 (lanelet 59)
- dangling reference to non existing intersection_area 777 (lanelet 53)

## Notes for reviewers

None.

## Effects on system behavior

None.


map/autoware_lanelet2_map_validator/src/validators/intersection/intersection_area_dangling_reference.cpp
TaikiYamada4, line 1: I think that the order
1. Collect all intersection_area polygons first
2. Scan through lanelets and find defected lanelets using `is_intersection_with_area`.

will be much efficient that we don't need to collect `intersection_with_area_lanelets`, but what do you think?

map/autoware_lanelet2_map_validator/test/src/test_intersection_area_dangling_reference.cpp
```diff
@@ -0,0 +1,83 @@
...
+protected:
+  void SetUp() override
+  {
+    // prepare `map_` and `loading_errors_`
+    // this class uses valid map
+    load_target_map("intersection/basic_intersection_area.osm");
+  }
+
+private:
+};
```
TaikiYamada4, line 47: I think only one class is enough and try loading the maps at the start of the test functions if there's no special reason.

You can write the validation subject on the second argument of `TEST_F`.

TaikiYamada4 (commented):
Thank you for the great work!
I wrote my opinions below so please check them out.

Aside these comments
- Could you write a brief document to explain what this validator can do?
- Could you add this to `autoware_requirement_set.json` like the following?

```json
  {
    "id": "vm-03-01",
    "validators": [
      {
        "name": "mapping.intersection.intersection_area_dangling_reference"
      }
    ]
  },
```

map/autoware_lanelet2_map_validator/src/validators/intersection/intersection_area_dangling_reference.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+lanelet::validation::Issues
+IntersectionAreaDanglingReferenceValidator::check_intersection_area_dangling_reference(
+  const lanelet::LaneletMap & map)
+{
+  // returns the VALUE of intersection_area key
+  auto is_intersection_with_area = [](const auto & lanelet) -> std::optional<lanelet::Id> {
+    if (lanelet.attributeOr("turn_direction", "none") == std::string("none")) {
+      return std::nullopt;
+    }
```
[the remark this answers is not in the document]
soblin: I think it is not good for a non-intersection lane to have "intersection_area" entry cause it's meaningless. And current implementation ignores that anyway. But I think we should sanitize that.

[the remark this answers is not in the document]
soblin: fixed in efe26926bd21d4c0cfc6c52f5238522ea684df7f

map/autoware_lanelet2_map_validator/src/validators/intersection/intersection_area_dangling_reference.cpp
soblin: fixed in efe26926bd21d4c0cfc6c52f5238522ea684df7f

map/autoware_lanelet2_map_validator/src/validators/intersection/intersection_area_dangling_reference.cpp
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  lanelet::validation::Issues issues;
+  for (const auto & [lanelet, intersection_area_id] : intersection_with_area_lanelets) {
+    if (intersection_area_ids.find(intersection_area_id) == intersection_area_ids.end()) {
+      issues.emplace_back(
+        lanelet::validation::Severity::Error, lanelet::validation::Primitive::Lanelet, lanelet.id(),
+        append_issue_code_prefix(
+          this->name(), 1,
+          "Lanelet of ID " + std::to_string(lanelet.id()) +
```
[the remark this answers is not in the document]
soblin: fixed in efe26926bd21d4c0cfc6c52f5238522ea684df7f

TaikiYamada4 (approved):
@soblin 
Since the functionality is perfect, I'll approve this PR :+1:
However I have some comments about the phrasing in the document so could you check them out? It also relates to the spelling check CI.

map/autoware_lanelet2_map_validator/docs/intersection/intersection_area_dangling_reference.md
```diff
@@ -0,0 +1,27 @@
...
+2. Check if intersection lanelet has "intersection_area" key and its value is contained in the above IDs
+
+The validator outputs the following issue with the corresponding ID of the primitive.
+
+| Issue Code                                         | Message                                                                                 | Severity | Primitive | Description                                                                        | Approach                                                             |
+| -------------------------------------------------- | --------------------------------------------------------------------------------------- | -------- | --------- | ---------------------------------------------------------------------------------- | -------------------------------------------------------------------- |
+| Intersection.IntersectionAreaDanglingReference-001 | "Dangling reference to non-existing intersection area of ID \<LANELET ID\> is detected" | Error    | Lanelet   | Lookup to `intersection_area` from the reporeted lanelet will cause runtime error. | Go to the reported lanelet and delete "intersection_area" key entry. |
```
soblin, line 20:
```suggestion
| Intersection.IntersectionAreaDanglingReference-001 | "Dangling reference to non-existing intersection area of ID \<LANELET ID\> is detected" | Error    | Lanelet   | The reported lanelet will cause a runtime error when attempting to look up the non-existent intersection_area. | Go to the reported lanelet and delete "intersection_area" key entry. |
```

map/autoware_lanelet2_map_validator/docs/intersection/intersection_area_dangling_reference.md
```diff
@@ -0,0 +1,27 @@
+# intersection_area_dangling_reference
+
+## Validator name
+
+mapping.intersection.intersection_area_dangling_reference
+
+## Feature
+
+This validator check whether each intersection lanelet(namely the lanelet with `turn_direction` property) has existing reference to `intersection_area` polygon. The countercase occurs when an existing intersection_area is deleted but its referrers are not updated.
```
soblin, line 9:
```suggestion
This validator checks whether each intersection lanelet(namely the lanelet with `turn_direction` property) has reference to existing `intersection_area` polygon. Such invalid case occurs when an intersection_area is deleted but its referrers are not updated.
```

PR #180, opened 2024-12-19
feat(lanelet2_map_validator): generation script for new validators

TaikiYamada4:
## Description

This PR adds a tool to `autoware_lanelet2_map_validator` that can generate necessary files to create a new validator which will be useful to contributors.

This PR does the following.
- Create a `template` directory and add templates for documents and test codes.
- Move `validator_template.cpp` and `validator_template.hpp` to `template`
- Create a python script `create_new_validator.py` that can generate necessary files for a new validator.
- Added information about how to use `create_new_validator.py` to ` how_to_contribute.md`.

## How was this PR tested?

Tested the following command at the `autoware_tools/map/autoware_lanelet2_map_validator` directory, and checked four files are generated correctly.
```bash
./template/create_new_validator.py --base_directory ./ \
--category_name intersection \
--code_name turn_direction_checking \
--class_name TurnDirectionValidator \
--validator_name mapping.intersection.turn_direction_tagging \
--check_function_name check_turn_direction
```

Then the output should be like
```bash
Create ./src/validators/intersection/turn_direction_checking.cpp
Create ./src/include/lanelet2_map_validator/validators/intersection/turn_direction_checking.hpp
Create ./test/src/intersection/test_turn_direction_checking.cpp
Create ./docs/intersection/turn_direction_checking.md
Process complete.
```

## Notes for reviewers

None.

## Effects on system behavior

None.


soblin (approved):
LGTM

`ros2 run lanelet2_map_validator create_new.py <args>` is also good

PR #188, opened 2025-01-07
fix(lanelet2_map_validator): restore missing intersection lane and removed unnecessary linestrings from intersection test maps

TaikiYamada4:
## Description

I found that one of the intersection lanes were removed by accident.
This PR restores that lane and removed unnecessary linestrings from intersection test maps.

Screenshot from 2025-01-07 12-26-24
The left figure is the current `basic_intersection_area.osm`. The red rectangle implies the missing lane.
The right figure is the one fixed in this PR.

## How was this PR tested?

Checked that following command runs without errors.
```bash
colcon test --packages-select autoware_lanelet2_map_validator --event-handlers console_cohesion+
```

## Notes for reviewers

The two maps below have large changes since they were not made from vector map builder but have been edited through vector map builder in this PR.
- `basic_intersection_area.osm`
- `intersection_area_with_wrong_orientation.osm`

## Effects on system behavior

None.

### DIFF
--- a/map/autoware_lanelet2_map_validator/test/data/map/intersection/basic_intersection_area.osm
+++ b/map/autoware_lanelet2_map_validator/test/data/map/intersection/basic_intersection_area.osm
@@ -2155,6 +2155,14 @@
     <tag k="type" v="line_thin" />
     <tag k="width" v="0.200" />
   </way>
+  <way id="197">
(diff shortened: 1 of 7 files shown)


soblin (approved):
Ths

PR #211, opened 2025-02-06
feat: add tools to collect TP values of PCD map segments

anhnv3991:
## Description
- The TPs resulting from NDT localization may vary on map, i.e. matching on different maps may yield very different TPs. This PR adds some tools to collects maps TPs. Those 'standard' TPs are used to verify localization results in other rosbags, and if some 'standard' TPs differ from the rosbags' TPs significantly, there may be a chance that the corresponding map segments have changed.
- Add scripts to collect TP values of PCD map segments and show the TPs on rviz
- Add a script to check compare TPs of a rosbag with the TPs of map segments and visualize the map areas where rosbag TPs differ from segment TPs significantly

## How was this PR tested?
- Tested on rinkai odaiba
- Visualization of segments TPs. The cyan parts are the segments with high TPs. The red parts are segments with low TPs (< 3.0)or no TPs.
-
Screenshot from 2025-01-28 10-27-34

- Results of checking rosbag TPs. Cyan parts are segments where rosbag TPs are not much different from map TPs. 
- 
Screenshot from 2025-01-28 10-39-10

## Notes for reviewers

This is a baseline implementation and the tools are still prototypes. TPs are collected by a super simple manner - segments will keep the max TP near them. We can compute the expected TP of a pose by querying the segments near the pose (search radius 20m), and averaging the segments' TPs. In general, this is not correct, as the scan may not be able to reach some segments. This should be fixed in future.

## Effects on system behavior

None.


map/autoware_tp_collector/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+
+find_package(autoware_cmake REQUIRED)
+autoware_package()
+
+# Enable support for C++17
+if (${CMAKE_VERSION} VERSION_LESS "3.1.0")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17")
+else ()
+  set(CMAKE_CXX_STANDARD 17)
+endif ()
```
SakodaShintaro, line 12: I think `autoware_cmake` uses C++17 by default.

https://github.com/autowarefoundation/autoware_cmake/blob/main/autoware_cmake/cmake/autoware_package.cmake#L18

map/autoware_tp_collector/README.md
```diff
@@ -0,0 +1,48 @@
+# autoware_pointcloud_merger
```
SakodaShintaro, line 1: autoware_pointcloud_merger?

map/autoware_tp_collector/package.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>autoware_tp_manager</name>
+  <version>0.1.0</version>
+  <description>A package for checking TP scores of NDT matching</description>
+  <maintainer email="anh.nguyen.2@tier4.jp">Anh Nguyen</maintainer>
```
SakodaShintaro, line 7: It is convenient to add some Localization&Mapping members as maintainers.

map/autoware_tp_collector/schema/pointcloud_merger.schema.json
```diff
@@ -0,0 +1,48 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Parameters for autoware point cloud merger node",
```
SakodaShintaro, line 3: point cloud merger?

map/autoware_tp_collector/CMakeLists.txt
SakodaShintaro: And since the main scripts are Python, is it needed?

PR #213, opened 2025-02-24
feat: add tools to collect and check average TPs of PCD maps

anhnv3991:
## Description
- Add a tool to collect average TPs of PCD map segments
- Add a tool to visualize the average TPs of PCD map segments to rviz2
- Add a tool to compare a rosbag TPs with a map's TPs and display the map areas where there are significant TP difference between the rosbag's TPs and the map's TPs

## How was this PR tested?
- Please see [README](https://github.com/autowarefoundation/autoware_tools/pull/213/files#diff-b7e8df8861d84dbe9c6828bfc3979bc2d369058e8b509f262b6c9ddf435354ac).
- Create TP file (.csv) with the old rosbag and pcd map by TP_collector
- Evaluate the new rosbag that has same route to old one and some changes of shape characteristic by TP_checker.
- Check the result that shows the segment where the shape characteristic changes.

## Notes for reviewers

None.

## Effects on system behavior

None.


Motsu-san:
@anhnv3991 Thank you for your pull-request!
I updated the test procedure of the description. If there is something wrong, please modify it.

Motsu-san (commented):
@anhnv3991 The change regarding pointcloud_divider is only to follow the latest in the README and the launch file, so there is no need to change the way we use the tool before and after the change, right?

map/autoware_tp_manager/package.xml
```diff
@@ -0,0 +1,28 @@
...
+  <maintainer email="anh.nguyen.2@tier4.jp">Anh Nguyen</maintainer>
+  <maintainer email="masahiro.sakamoto@tier4.jp">Masahiro Sakamoto</maintainer>
+  <license>Apache License 2.0</license>
+
+  <author email="anh.nguyen.2@tier4.jp">Anh Nguyen</author>
+  <author email="masahiro.sakamoto@tier4.jp">Masahiro Sakamoto</author>
+
+  <buildtool_depend>ament_cmake_auto</buildtool_depend>
+  <buildtool_depend>autoware_cmake</buildtool_depend>
+  <buildtool_depend>tier4_debug_msgs</buildtool_depend>
```
Motsu-san, line 19: I would like to have it made with autoware_internal_debug_msgs as much as possible, are there any restrictions?

map/autoware_tp_manager/README.md
```diff
@@ -0,0 +1,85 @@
+# autoware_tp_manager
+
+Here are some tools for collecting average TPs of PCD maps. Currently, we consider the decrease of TPs as a sign of map decay. However, we don't know what TPs are 'abnormal', e.g. in some areas the TPs range around 2.0 ~ 3.0, while in others TPs float around 5.0. This package provides some tools to check it, including:
+
+- TP collector: collect the average TPs per segment of a PCD map
+- TP checker: compare a rosbag's TPs with a map's TPs and highlight the map areas where the rosbag's TPs differ significantly from the map's TPs.
+
```
Motsu-san, line 7: Could you add the python requirements?

Motsu-san (changes requested):

anhnv3991:
@Motsu-san 
> @anhnv3991 The change regarding pointcloud_divider is only to follow the latest in the README and the launch file, so there is no need to change the way we use the tool before and after the change, right?

Yes, we can still use pointcloud_divider like before. I just move parameters to the launch file to give us more flexibility to interract with this pointcloud_divider from the other programs.

map/autoware_tp_manager/package.xml
anhnv3991: @Motsu-san Thank you for your comment.
I replaced tier4_debug_msgs by autoware_internal_debug_msgs.

map/autoware_tp_manager/README.md
anhnv3991: @Motsu-san You mean something like this or a requirements.txt file?
`Here are some Python tools for collecting average TPs of PCD maps. Currently, we consider the decrease of TPs as a sign of map decay...`

anhnv3991:
@TaikiYamada4 
> @anhnv3991 The LICENSE file says BSD 3-Clause but the python scripts say Apache-2.0. Is that OK?

Thank you for checking. I replaced the LICENSE file by Apache-2.0.

TaikiYamada4:
@anhnv3991 
> Thank you for checking. I replaced the LICENSE file by Apache-2.0.

If that's the case, the `autoware_tools` repository is basically Apache-2.0 so maybe the LICENSE file itself is not necessary.

map/autoware_tp_manager/README.md
Motsu-san: @anhnv3991 If there is other requirements than autoware's ones, I would like to see them in README or requirements.txt file.

anhnv3991: @Motsu-san I think no other requirement is needed.

Motsu-san (approved):
[LGTMeow](https://lgtmeow.com)

PR #225, opened 2025-03-19
fix(autoware_static_centerline_generator): changes to support modifications made to autoware.core

PanConChicharron:
## Description
Makes appropriate changes to support new changes made to autoware.core
## Notes for reviewers

None.

## Effects on system behavior

None.


xmfcx (approved):
It compiles now, thx for the update 👍

PR #239, opened 2025-04-24
feat(build_depends.repos): use fixed version for autoware and universe repository

Kazunori-Nakajima:
## Description

- Some packages related to Autoware are version-locked

- Some Autoware-related packages are branch specific
  This situation makes it difficult to pass builds when changes are made that depend on multiple repositories.

To solve this, revert all Autoware and universe packages to branches.

## Notes for reviewers

None.

## Effects on system behavior

None.

### DIFF
--- a/build_depends.repos
+++ b/build_depends.repos
@@ -3,7 +3,7 @@ repositories:
   autoware/autoware_cmake:
     type: git
     url: https://github.com/autowarefoundation/autoware_cmake.git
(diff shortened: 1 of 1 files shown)


kosuke55 (commented):
LGTM to prevent build fail like
https://github.com/autowarefoundation/autoware_tools/pull/238

kosuke55 (approved):

PR #241, opened 2025-04-29
feat(static_centerline_generator): visualize iterative planning results and introduce glog

takayuki5168:
## Description

- 1. introduced glog 
    - build_depends.repos is updated to clone tier4/glog.
- 2. added a visualization of the iterative path planning
- 3. fixed several bugs
    - error handling when the centerline's size on the lanelet2 map is less than or equal to 2.
    - added an orientation check in addition to the position check to handle the overlapped lanes

2 can show us the following iterative visualization
- yellow: smoothed trajectory
- blue: optimized trajectory

https://github.com/user-attachments/assets/e2b93f97-fbc5-49fa-9aaf-39c33c2d36ec

## How was this PR tested?
Unit test passed
## Notes for reviewers

None.

## Effects on system behavior

None.

### DIFF
--- a/build_depends.repos
+++ b/build_depends.repos
@@ -29,7 +29,7 @@ repositories:
   core/autoware_lanelet2_extension:
     type: git
     url: https://github.com/autowarefoundation/autoware_lanelet2_extension.git
-    version: main
+    version: 0.7.0
   # universe
   universe/autoware.universe:
     type: git
--- a/build_depends.repos
+++ b/build_depends.repos
@@ -43,3 +43,7 @@ repositories:
     type: git
     url: https://github.com/MORAI-Autonomous/MORAI-ROS2_morai_msgs.git
     version: main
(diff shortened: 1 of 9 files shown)


build_depends.repos
```diff
@@ -29,7 +29,7 @@ repositories:
   core/autoware_lanelet2_extension:
     type: git
     url: https://github.com/autowarefoundation/autoware_lanelet2_extension.git
-    version: main
```
takayuki5168, line 32: The build error happens with the main branch.
The same version in the autoware.universe will be used with this change.
https://github.com/autowarefoundation/autoware_universe/blob/8f64d8eaee6bf73358b38309951b8263968be830/build_depends_humble.repos#L14

Kazunori-Nakajima (approved):
Thank you very much!! LGTM!

PR #243, opened 2025-04-30
feat(planning/perception_reproducer): publish initial ego pose

maxime-clem:
## Description

- Define common function to publish an initial pose.
- new feature: with `perception_reproducer.py`, if the ego pose is not set, publish the first recorded one.

## How was this PR tested?

Psim

## Notes for reviewers

None.

## Effects on system behavior

None.


maxime-clem:
A similar feature is already available with the `--publish-route` option.

PR #252, opened 2025-05-14
feat: enable pcd divider to generate a metadat file

anhnv3991:
## Description
- Enable the point cloud divider to generate a metadata file of map segments, instead of actually dividing the PCD map. This is helpful when we work with tiny segmentation resolution without needing the actual PCD files. In such cases, dividing the PCD map is super slow, but generating a metadata file is very fast.

## How was this PR tested?
- I tested this on the large Odaiba rinkai map, trying to divide the map to tiny 2.0x2.0m segments.
- Run this command 
```ros2 launch autoware_pointcloud_divider pointcloud_divider.launch.xml input_pcd_or_dir:=<map_path> output_pcd_dir:=<output_folder> grid_size_x:=2.0 grid_size_y:=2.0 metadata_generate:=true```
- A metadata file was generated at the <output_folder>

## Notes for reviewers

None.

## Effects on system behavior

None.

### DIFF
--- a/map/autoware_pointcloud_divider/src/pcd_divider.cpp
+++ b/map/autoware_pointcloud_divider/src/pcd_divider.cpp
@@ -52,6 +52,7 @@
 #include <list>
 #include <memory>
 #include <string>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 
@@ -101,13 +102,19 @@ std::vector<std::string> PCDDivider<PointT>::discoverPCDs(const std::string & in
 }
 
 template <class PointT>
-void PCDDivider<PointT>::run()
+void PCDDivider<PointT>::run(bool meta_gen)
 {
   // Discover PCD files
   auto pcd_list = discoverPCDs(input_pcd_or_dir_);
 
   // Process pcd files
-  run(pcd_list);
+  if (meta_gen) {
+    // Only generate a metadata file
+    meta_generator(pcd_list);
+  } else {
+    // Do the segmentation
+    run(pcd_list);
+  }
 }
 
 template <class PointT>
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: map/autoware_pointcloud_divider/src/pcd_divider.cpp [hunk 1]
```diff
@@ -164,15 +171,48 @@ void PCDDivider<PointT>::checkOutputDirectoryValidity()
...
+  const std::string & pcd_name, bool load_all)
 {
   if (pcd_name != reader_.get_path()) {
     reader_.setInput(pcd_name);
   }
 
   PclCloudPtr cloud_ptr(new PclCloudType);
 
-  reader_.readABlock(*cloud_ptr);
+  if (load_all) {
```
YamatoAndo, line 183: Why is the `load_all` branch necessary?
In this program, it is always set to false, so this branch is never executed.
Having unused code increases maintenance cost, so please remove it.

YamatoAndo:
This is a package intended for dividing map point clouds,
so I don't understand the point of adding a flag that disables that functionality.

YamatoAndo:
You should consider refactoring the code into a library so that both `pcd divider` and `tp manager` can generate metadata.

map/autoware_pointcloud_divider/src/pcd_divider.cpp
anhnv3991: @anhnv3991 I removed it.

anhnv3991:
@YamatoAndo 
> This is a package intended for dividing map point clouds, so I don't understand the point of adding a flag that disables that functionality.

I agree that the point cloud divider should divide a map into actuall PCD files. But we don't always want to process maps at point-level. If we examine maps at the segment-level, the metadata is enough. The TP manager now is checking Tps of map segments, so need to actually divide map points.

I want to keep all the operations related to hashing points, finding segments, and distributing points to segments in a single package. So adding a parameter to generate a metadata file without dividing points is enough, imo.

anhnv3991:
@YamatoAndo 
> You should consider refactoring the code into a library so that both `pcd divider` and `tp manager` can generate metadata.

I will consider this when there are more than two packages that call point cloud divider's functionality. Currently, adding one parameter and call the pcd divider via a python's call is simpler.

YamatoAndo:
> Currently, adding one parameter and call the pcd divider via a python's call is simpler.

It's best to stop using it like that, as it makes no sense.

PR #262, opened 2025-06-18
feat(trajectory_visualizer): add visualization of X, Y, and Yaw values

maxime-clem:
## Notes for reviewers

None.

## Effects on system behavior

None.


kyoichi-sugahara (approved):
image
It worked well in my environment! thanks.

PR #274, opened 2025-07-17
fix(static_centerline_generator): inside centerline start point in start lanelet

Kazunori-Nakajima:
## Description

Fixed a bug where start_lanelet was on a duplicate route and the centerline could not be embedded when using the duplicate route.

### Explanation of cause

Assume there is a map with the following overlapping routes.

image

When creating a centerline route with path_generator, the default path_length.backward = 0.5m was used.
Therefore, a centerline from the yellow point to GOAL was created internally.
The process then checks to see if the start_point of the centerline is in a lanelet in the path.

In the case of this map, the start_point of the centerline is contained in the goal_lanelet instead of the start_lanelet.
This is not expected and is a bug.

image

### Corresponding method

Fixed so that centerline is created from within start_lanelet by passing path_length.backward = -0.01 m.

image

## How was this PR tested?

[use latest map(TIER IV INTERNAL LINK)](https://evaluation.tier4.jp/evaluation/maps/1130?project_id=autoware_dev)

```sh
ros2 launch autoware_static_centerline_generator static_centerline_generator.launch.xml centerline_source:='optimization_trajectory_base' mode:='AUTO' lanelet2_input_file_path:={local_map_path} vehicle_model:='cargo_transport' goal_method:='path_generator' start_lanelet_id:='9640' end_lanelet_id:='11468' end_pose:='[93880.56671685721, 10499.697589653153, 150.1852857142857, 0.0, 0.0, -0.8777981653421517, 0.47903066803906463]'
```

https://github.com/user-attachments/assets/b9860688-5b57-4bb9-b5b8-d4ca765a707a

## Notes for reviewers

None.

## Effects on system behavior

None.


Kazunori-Nakajima:
@takayuki5168 
Could you please review this pr :bowing_man:

PR #279, opened 2025-07-31
Fix parameter estimator

Kazunori-Nakajima:
## Description

- Updated `vehicle_launch` → `tier4_vehicle_launch`
- Updated `map_launch` → `tier4_map_launch`
- Removed dependency on the deleted `dynamic_object_visualization` package

## Notes for reviewers

None.

## Effects on system behavior

None.


Kazunori-Nakajima:
We will handle it https://github.com/autowarefoundation/autoware_tools/pull/280

Issue #281, opened 2025-07-31
Fix flaky test: test_static_centerline_generator_path_generator_case3_launch.test.py


xmfcx:
### Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I'm convinced that this is not my fault but a bug.

### Description

Sort of flaky test 🤷‍♂️

Fails on the test from the universe side:
- https://github.com/autowarefoundation/autoware_universe/actions/runs/16647774294/job/47114699646?pr=11091#step:17:17150

### Steps to reproduce

```console
cd autoware
git checkout main
rm build/ log/ install/ src/* # This will delete the contents of src folder!
vcs import src < autoware.repos
vcs import src < autoware-nightly.repos
vcs import src < tools-nightly.repos

colcon build --symlink-install --event-handlers=console_cohesion+ --cmake-args -DCMAKE_BUILD_TYPE=RelWithDebInfo  --packages-up-to autoware_static_centerline_generator

colcon test --event-handlers console_cohesion+ --packages-select autoware_static_centerline_generator
```

https://gist.github.com/xmfcx/0b40728f0379480bfe1c706e859c307b

I have seen this both fail and succeed on my machine. But even when it succeeds, like in the logs in the gist, it will say it failed.
```
10: [main-3] 1. Footprints inside Lanelets:
10: [main-3]  The generated centerline is outside the lanelet. (actual:0[m] <= threshold:0[m])
10: [main-3]   Failed.
10: [main-3] 2. Curvature:
10: [main-3]   The generated centerline has no high steer angle. (estimated:37.3588[deg] < threshold:44.9772[deg])
10: [main-3]   Passed.
10: [main-3] 
10: [main-3] Result:
10: [main-3]   Failed!
```

Similarly this happened in the PR that introduced this test:
- https://github.com/autowarefoundation/autoware_tools/actions/runs/15003669689/job/42156802818#step:9:762

**Related PRs:**
- https://github.com/autowarefoundation/autoware_tools/pull/250#issuecomment-3140003653
- https://github.com/autowarefoundation/autoware_universe/pull/11091#issuecomment-3140118624

PR #282, opened 2025-07-31
fix(autoware_static_centerline_generator): increase test timeouts

xmfcx:
## Description

- Aims to resolve: https://github.com/autowarefoundation/autoware_tools/issues/281

## Notes for reviewers

None.

## Effects on system behavior

None.


xmfcx:
- I decided to continue this fix here: https://github.com/autowarefoundation/autoware_tools/pull/283 because some other CI related things need to be fixed too. Like autoware_lanelet2_extension version update is needed to make https://github.com/autowarefoundation/autoware_tools/actions/runs/16657161800/job/47144972860?pr=282 pass.

PR #284, opened 2025-08-01
fix(static_centerline_generator): set unit test timeout to 60s

kosuke55:
## Description

currently sometimes unit test fails due 30s timeout.
so increase value to 60s

https://github.com/autowarefoundation/autoware_universe/actions/runs/16672616115/job/47192430950?pr=11101

## How was this PR tested?

not tested 

## Notes for reviewers

None.

## Effects on system behavior

None.


kosuke55:
@xmfcx 
thanks a lot!!!!!!!!!!!!!!!!!!!!

PR #285, opened 2025-08-04
fix(autoware_static_centerline_generator): fix test dependency

mitsudome-r:
## Description
This PR fixes dependency for autoware_static_centerline_generator.
The package uses Qt5 for execution, but it does not have the library as dependency.

This is causing a failure in build-and-test-differential-above in Autoware Core.
https://github.com/autowarefoundation/autoware_core/actions/runs/16671444714/job/47188492044?pr=593
image

## How was this PR tested?
The test passes in CI. 

## Notes for reviewers

None.

## Effects on system behavior

None.


takayuki5168 (approved):
Thank you

PR #297, opened 2025-08-26
feat(autoware_localization_evaluation_scripts): create masks for each criterion

TaikiYamada4:
## Description

Currently, autoware_localization_evaluation_scripts can only provide the success/failure result by taking the AND condition of all criteria.This PR enables autoware_localization_evaluation_scripts to toggle whether to use the criterion or not for the final summary.

The mask is expected to be provided from the scenario file by adding a condition like the following.
Note that passing a scenario file is optional and the script will manage the masks as true if no scenario file is given OR there is no `OverallCriteriaMask` condition in the scenario file. So after this PR is merged, the existing evaluation through driving_log_replayer doesn't change at all.

Read the modified `README.md` for further information.

```yaml
Evaluation:
  Conditions:
    OverallCriteriaMask:
      mean_relative_position: false
      mean_relative_angle: false
      mean_relative_linear_velocity: true
      mean_relative_angular_velocity: true
      mean_relative_acceleration: false
      diagnostics_not_ok_rate: false
```

## How was this PR tested?

Tested by the following procedure.

1. Get a random scenario file and its evaluation result and from the driving_log_replayer.
2. Add a `OverallCriteriaMask` block to the scenario file above, OR you can just save this as a new yaml file and treat it as a scenario file which will work only for testing this PR.
3. Build `autoware_localization_evaluation_scripts` if you haven't.
4. Execute the following

```bash
ros2 run autoware_localization_evaluation_scripts analyze_rosbags_parallel.py \
/path/to/driving_log_replayer/result \
--save_dir_relative result_archive \
--topic_reference /localization/reference_kinematic_state \
--scenario_file scenario.yml
```

## Notes for reviewers

None.

## Effects on system behavior

Skip criterion if it is set as `false` in the scenario file.

### DIFF
--- a/localization/autoware_localization_evaluation_scripts/scripts/analyze_rosbags_parallel.py
+++ b/localization/autoware_localization_evaluation_scripts/scripts/analyze_rosbags_parallel.py
@@ -46,13 +46,10 @@ def load_overall_criteria_mask(yaml_path: Path) -> OverallCriteriaMask:
             data = yaml.safe_load(f)
         conditions = data["Evaluation"]["Conditions"]
     except Exception:
-        print("Cannot load scenario file")
         return OverallCriteriaMask()
 
     # If mask exists -> merge with defaults, else return defaults
     mask = conditions.get("OverallCriteriaMask", {})
-    if mask is {}:
-        print("mask is empty")
     return OverallCriteriaMask(**{**OverallCriteriaMask().__dict__, **mask})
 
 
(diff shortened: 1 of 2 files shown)


localization/autoware_localization_evaluation_scripts/scripts/analyze_rosbags_parallel.py
```diff
@@ -24,18 +36,49 @@ def parse_args() -> argparse.Namespace:
...
     return parser.parse_args()
 
 
+def load_overall_criteria_mask(yaml_path: Path) -> OverallCriteriaMask:
+    try:
+        with open(yaml_path, "r") as f:
+            data = yaml.safe_load(f)
+        conditions = data["Evaluation"]["Conditions"]
+    except Exception:
+        return OverallCriteriaMask()
```
SakodaShintaro, line 49: It's a little confusing that nothing is displayed when loading fails, but I don't think there's any need to fix it.

SakodaShintaro (approved):

PR #307, opened 2025-09-29
feat(trajectory_visualizer): sort topic list

Kotakku:
## Description
This PR contains updates for trajectory_visualizer. The topic list is now sorted by node and highlights key topics.

- Grouping and Order: Topics are now grouped and separated by their publishing node (or module) name. The display flow is top-down, following the data trajectory.
- Key Topic Highlighting: Topics that represent key connections (i.e., those connected to the next node or module) are visually highlighted in blue for easy identification of the data flow path.

image

## Notes for reviewers

None.

## Effects on system behavior

None.


maxime-clem (commented):
I tested the PR and the feature to pre-select some topics with config files no longer works. :cry:

kyoichi-sugahara (approved):

PR #321, opened 2025-12-05
feat(perception_replayer): cpp implementation of perception replayer

Kotakku:
## Description
The Python implementations of `perception_replayer` and `perception_reproducer` are bottlenecked by msg copy and `rclpy::deserialize`, which slows startup and playback. To address this, both tools were reimplemented in C++.

The measured startup times for several rosbags comparing the Python and C++ implementations are shown below.

| Case                                                          | Python version | C++ version |
| ------------------------------------------------------------- | -------------- | ----------- |
| 1-minute rosbag including point clouds and camera images      | 20-30 s        | under 5 s   |
| 1-minute rosbag with a duplicated `traffic_light` topic (1)   | 60-90 s        | under 10 s  |
| 1-minute rosbag without point clouds and camera images (2)    | 15-25 s (3)    | under 1 s   |

1: Rosbag that unintentionally contains duplicated V2I information.  
2: Rosbag storing only `kinematic_state`, `object`, and `traffic_signals`.  
3: For the Python version, msg copy dominates more than db3 file deserialization, so the time does not change much.

After this PR, the execution commands change as follows:
```bash
ros2 run planning_debug_tools perception_reproducer -b <bag-file>
```

```bash
ros2 run planning_debug_tools perception_replayer -b <bag-file>
```

## Notes for reviewers

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/planning_debug_tools/src/perception_replayer/perception_reproducer.cpp
+++ b/planning/planning_debug_tools/src/perception_replayer/perception_reproducer.cpp
@@ -214,7 +214,7 @@ void PerceptionReproducer::on_timer()
     publish_topics_at_timestamp_with_coordinate_conversion(
       bag_timestamp.value(), current_timestamp, ego_odom, param_.noise && repeat_flag);
   } else {
-    RCLCPP_WARN(get_logger(), "No valid bag timestamp to publish.");
+    RCLCPP_WARN_THROTTLE(get_logger(), *get_clock(), 3000, "No valid bag timestamp to publish.");
   }
 }
 


kyoichi-sugahara (approved):
Super thanks!!!
Confirmed the new versions work in my environment with following option and feel big speed up :heart: 
```
ros2 run planning_debug_tools perception_reproducer -b <bag-file>
ros2 run planning_debug_tools perception_reproducer -b <bag-file> -p
ros2 run planning_debug_tools perception_replayer -b <bag-file>
```

So LGTM from my perspective, thanks!!!!!

maxime-clem (approved):
Awesome work !

The only real issue I found is that the button of the replayer to adjust the replay rate do not work. This can be fixed in a future PR.

PR #329, opened 2025-12-24
feat(reaction_analyzer): move package from autoware_universe

isamu-takagi:
## Description

**Use merge commit to keep history**

To remove the dependency on autoware_launch from autoware_universe, move reaction_analyzer from autoware_universe.
https://github.com/autowarefoundation/autoware_universe/pull/11807

## How was this PR tested?

Check build.

## Notes for reviewers

None.

## Effects on system behavior

None.


mitsudome-r (approved):

isamu-takagi:
Thank you. I'll merge this PR with merge commit to keep history.

system/reaction_analyzer/package.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>reaction_analyzer</name>
+  <version>0.49.0</version>
```
mitsudome-r: It would be nice if we can match the package version with the rest of the packages.
We should also consider resetting the CHANGELOG.rst as well.

PR #336, opened 2026-01-13
feat(autoware_route_history): save, load, and manage pose-to-pose routes in the Autoware planning tool

lewisjsmith:
# Route history and replay for the planning tool

## Overview

This package provides an interactive RViz panel for saving poses, enabling the quick (re)generation of routes for simulation and testing.

## Motivation

Repeatedly testing a specific route is cumbersome due to the need to replot to re-run. This can be done manually through "Add goal pose" and other toolbar tools, but the use of the GUI is prone to inaccuracy. An alternative is to publish to the pose topics. This RViz panel aims to streamline this process to record and manage multiple routes.

## Quick Start

```sh
cd ~/autoware_tools/planning
colcon build --packages-select autoware_route_history
source install/setup.bash
```

After launching the planning tool the RouteHistory panel will be discoverable.
Panels > Add New Panel > RouteHistoryPanel > OK.

## Output Files

- **`~/.ros/route_history.yaml`** - Saved routes are paired with uuids and written to a YAML file. A new file path can be chosen through the "Load save file" button in the running RViz panel.

## Parameters

- **`save_file_path`** - Holds the path string associated with the location of the save file.

## Future Work

- **Groups** - group routes for sequential, automatic simulation.
- **Terminal access to features** - easy-to-use terminal commands for managing tests.

### DIFF
--- a/planning/autoware_route_history/include/route_history/node_logic.hpp
+++ b/planning/autoware_route_history/include/route_history/node_logic.hpp
@@ -65,7 +65,9 @@ using uuid_route_map = std::unordered_map<std::string, NamedRoute>;
 class NodeLogic
 {
 public:
-  explicit NodeLogic(const rclcpp::Node::SharedPtr & node) : node_(node)
+  // clang-format off
+  explicit NodeLogic(const rclcpp::Node::SharedPtr & node)
+  : node_(node)  // clang-format on
   {
     if (!node_->has_parameter("save_file_path")) {
       node_->declare_parameter("save_file_path", "~/.ros/route_history.yaml");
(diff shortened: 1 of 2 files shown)


lewisjsmith:
@mojomex

planning/pose_replay/pose_replay_interfaces/srv/DeleteRoute.srv
```diff
@@ -0,0 +1,3 @@
+string uuid
+---
+string uuid
```
mojomex: For this, and also `SetName` and `SetRoute`, do you care about the UUID in the response? From what I understand, it would just be the same as the UUID in the request and thus redundant.
If you just want a mechanism for signaling success/error, a `bool` would be more appropriate.

planning/pose_replay/pose_replay_interfaces/srv/GetUuidRoute.srv
mojomex: Given that `GetUuidRoutes` returns a `UuidRoutes` type, I would expect this service to return a `UuidRoute` message. Maybe change the name to something more descriptive, like `GetRouteDetails` or so.

planning/pose_replay/pose_replay_interfaces/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+cmake_minimum_required(VERSION 3.8)
+project(pose_replay_interfaces)
+
```
mojomex: While not strictly necessary here, it's good practice to explicitly set the C++ standard in all build scripts.

PR #398, opened 2026-04-13
feat(planning_data_analyzer): parallelize open-loop evaluation with  multiple EPDMS horizons

beomseok-kimm:
## Description

This PR parallelizes open-loop evaluation in `autoware_planning_data_analyzer` and ports the implementation onto the current `autowarefoundation/autoware_tools` `main` base.

Included in this PR:

- parallel trajectory evaluation with worker threads for open-loop metrics
- parallel phase split for base metrics and extended comfort / synthetic EPDMS horizon evaluation
- preservation of bag-output equivalence by restoring merge-based final bag generation

The bag finalization path keeps the final output behavior aligned with the previous sequential implementation:

- intermediate bag -> analyzer temp bag -> merged final bag
- no build artifacts are included in this PR

## How was this PR tested?

 ran the full offline pipeline and compared the artifact created before the change this time with the current output:

Validation results from the full pipeline:

- output bag size matched: `56G` vs `56G`
- EPDMS-related aggregate values matched exactly between reference and rerun

Observed analyzer runtime from the successful rerun:

- total analyzer runtime: about `1431.82 s` (`23m 51.8s`)
- open-loop evaluation core: about `1211.38 s` (`20m 11.4s`)
- parallel phase 1: about `527.44 s`
- parallel phase 2: about `611.36 s`

### DIFF
--- a/planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
+++ b/planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
@@ -14,10 +14,17 @@
 
 #include "open_loop_evaluator.hpp"
 
+#include "metrics/drivable_area_compliance.hpp"
 #include "metrics/epdms_aggregation.hpp"
+#include "metrics/history_comfort.hpp"
 #include "metrics/metric_utils.hpp"
+#include "metrics/no_at_fault_collision.hpp"
+#include "metrics/traffic_light_compliance.hpp"
 #include "metrics/trajectory_metrics.hpp"
+#include "metrics/ttc_within_bound.hpp"
 
+#include <autoware/lanelet2_utils/geometry.hpp>
+#include <autoware/lanelet2_utils/intersection.hpp>
 #include <autoware/motion_utils/trajectory/conversion.hpp>
 #include <autoware/motion_utils/trajectory/trajectory.hpp>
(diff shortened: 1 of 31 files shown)


planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
```diff
@@ -219,126 +661,261 @@ void OpenLoopEvaluator::evaluate(
...
+    evaluation_data_list.size(), worker_count);
+  RCLCPP_INFO(logger_, "Parallel phase 1 started: base trajectory metrics");
+
+  std::atomic<std::size_t> next_index{0U};
+  std::atomic<std::size_t> phase1_processed{0U};
+  std::atomic<bool> stop_requested{false};
+  std::exception_ptr worker_exception;
+  std::mutex worker_exception_mutex;
+
+  auto worker = [&]() {
```
go-sakayori, line 693: Unless RouteHandler is explicitly guaranteed to be thread-safe for concurrent read access, this introduces a real risk of data races or non-deterministic evaluation results. Since the main value of this PR is the new parallelism, I think we should either:

- document and prove that RouteHandler is safe to share across threads here, or
- avoid sharing it across workers, or
- serialize the RouteHandler access behind a lock.

PR #400, opened 2026-04-15
feat(planning_data_analyzer): remove override scene related source and script

go-sakayori:
## Description
Removed OR scene related sources, since it can be evaluated with the open loop simulation

## How was this PR tested?
Colcon build

## Notes for reviewers
None.

## Effects on system behavior
None.


beomseok-kimm (approved):
LGTM

PR #401, opened 2026-04-17
feat(planning_data_analyzer): parallelize open-loop evaluation and reduce metric-path overhead

beomseok-kimm:
## Description

This PR contains two planning data analyzer changes on top of `awf-latest`:

1. CPU parallelization for open-loop evaluation in `OpenLoopEvaluator`
2. metric-path optimization for the hottest calculation helpers

The first commit parallelizes the base metric evaluation and the extended-comfort / synthetic-EPDMS finalization phases while keeping the output phase sequential.

The second commit reduces repeated work in the metric hotspot path by:
- caching highest-confidence predicted-path metadata once per object during metric evaluation
- reusing the ego vehicle local footprint template instead of rebuilding the same base footprint repeatedly

### Run time comparison
Rosbag:  filtered bag (only required topics)
CPU: Intel(R) Core(TM) i9-14900K (32 cores)

- Parallelization optimization:
Total process time from **260sec** to **38sec** (`6.8x faster`)
- Repeated logic optimization
Metric compuation phase improved by **0.60 s**, about **6.7%**

For the compared filtered run, EPDMS and related aggregate metric outputs remained unchanged.

###  Time comparison and topic reproduction confirmation
- comparing analyzer phase timings before/after the metric optimization on a filtered evaluation run
- comparing EPDMS-related aggregate outputs before/after and confirming they were unchanged in that comparison

### Evaluator run
https://evaluation.tier4.jp/evaluation/reports/e76a3b8f-fa57-57c8-9a62-ceca1a9b41bb?project_id=x2_dev

## Notes for reviewers

- The PR is intentionally split into two commits:
  1. parallel open-loop evaluation
  2. metric-path optimization

## Effects on system behavior

No intended metric behavior change. The goal is to reduce analyzer runtime while keeping EPDMS and related metric outputs unchanged.


go-sakayori:
@beomseok-kimm 
Can you rebase properly? There are so many commits unrelated to the PR

beomseok-kimm:
> @beomseok-kimm Can you rebase properly? There are so many commits unrelated to the PR

Done!! review please

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
```diff
@@ -219,124 +293,244 @@ void OpenLoopEvaluator::evaluate(
...
+  if (worker_exception) {
+    std::rethrow_exception(worker_exception);
+  }
+
+  RCLCPP_INFO(
+    logger_, "Parallel phase 2 finished: processed %zu trajectory samples",
+    phase2_processed.load());
+  RCLCPP_INFO(logger_, "Sequential output phase started: bag writing and summary aggregation");
+
+  for (std::size_t i = 0; i < evaluation_data_list.size(); ++i) {
```
go-sakayori: This keeps all per-sample results in `parallel_results` and then copies them into `metrics_list_`, `trajectory_point_metrics_list_`, `human_filter_metrics_list_`, and `synthetic_epdms_metrics_list_`.

Because these result objects contain large per-trajectory vectors, this likely doubles peak memory usage during evaluation and adds extra copy overhead for long bags. Could we move the results out of `parallel_results`, or pre-size the member vectors and let workers write to indexed slots directly?

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
```diff
@@ -219,124 +293,244 @@ void OpenLoopEvaluator::evaluate(
...
+      }
+
+      const auto i = next_index.fetch_add(1U);
+      if (i >= evaluation_data_list.size()) {
+        return;
+      }
+
+      try {
+        const auto & eval_data = evaluation_data_list.at(i);
+        auto trajectory_metrics = metrics::calculate_trajectory_point_metrics(
```
go-sakayori, line 351: Question: the parallel workers now call metric calculations concurrently while sharing the same `route_handler_` instance.

I understand the route handler is initialized from the bag before evaluation and should not be mutated during this loop, so this may be fine. Could you confirm that the `RouteHandler` / Lanelet2 calls used here are safe for concurrent read-only access, or mention the validation you used for this? For example, these paths call methods such as `getRoadLaneletsAtPose()`, `getClosestLaneletWithinRoute()`, `isRouteLanelet()`, and `getLaneletMapPtr()` from multiple worker threads.

go-sakayori:
@beomseok-kimm 
Could you also add the link to the TIER IV internal test in the PR description?

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
beomseok-kimm: Addressed in f8ae837. Optimized by pre-allocating member vectors and using direct indexed writing in workers, eliminating the large temporary `parallel_results` vector and avoiding extra copies/doubled peak memory.

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
beomseok-kimm: Addressed in f8ae837. Lanelet2 maps in Autoware are immutable after loading, making read-only access generally safe. To ensure thread-safety against Lanelet2's internal lazy-loading/caching (e.g., spatial indices, centerlines), I added a `priming` step in the sequential phase to build these caches before entering the parallel loop.

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
```diff
@@ -212,142 +286,262 @@ void OpenLoopEvaluator::evaluate(
...
-          *previous_trajectory, *current_trajectory, extended_comfort_parameters_);
-        metrics.extended_comfort = extended_comfort_result.score;
-        metrics.extended_comfort_available = extended_comfort_result.available;
-        metrics.extended_comfort_reason = extended_comfort_result.reason;
+  // Prime RouteHandler to build internal caches/indices in a single-threaded context.
+  // This ensures thread-safety for subsequent parallel read-only access.
+  if (route_handler_ && route_handler_->isHandlerReady()) {
+    (void)route_handler_->getLaneletMapPtr();
+    if (!evaluation_data_list.empty()) {
+      (void)route_handler_->getRoadLaneletsAtPose(evaluation_data_list.front().synchronized_data->trajectory->points.front().pose);
```
go-sakayori: This can call `points.front()` on an empty trajectory.

`prepare_evaluation_data()` keeps `num_pts <= 1` trajectories as failure frames, so `evaluation_data_list.front().synchronized_data->trajectory->points` may be empty here. Could we guard this with `!points.empty()` or prime using the first sample that has at least one point?

go-sakayori (changes requested):

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
beomseok-kimm: Necessary fix. `prepare_evaluation_data` can return entries with empty trajectories for failure frames (0 or 1 points). Calling `points.front()` on these without a guard would cause a crash. I have added a safety guard using `std::find_if` to find the first non-empty trajectory before priming the `RouteHandler`

go-sakayori (approved):

PR #412, opened 2026-04-23
feat(trajectory_kinematics_panel): trajectory_kinematics_panel upgrade

Owen-Liuyuxuan:
## Description

Improve usability and performance of trajectory kinematics panel.

- Correctly handle topics not being published yet (at initialization)
- Avoid automatically refresh/subscribe on ROS2 topic list to improve performance (especially on vehicle).
- Improve UI designs
   - Only allow dynamic range, cleaned up unnecessary options on panels. So the panel area can be fully utilized.
   - Adjust the Refresh button and width composition, now the panel can be shrink to be smaller, more flexible on vehicle screen with 1080p screen (developer screen is 2K so I missed this problem before)
   - Limit the height range of series list.

I  am planning to have two independent PRs to add the feature on AWF & tier4 side
   

image

## Notes for reviewers

None.

## Effects on system behavior

None.

### DIFF
--- a/common/autoware_trajectory_kinematics_rviz_plugin/src/trajectory_series_manager.cpp
+++ b/common/autoware_trajectory_kinematics_rviz_plugin/src/trajectory_series_manager.cpp
@@ -104,6 +104,22 @@ void TrajectorySeriesManager::setTopicConfigs(
     std::lock_guard<std::mutex> lock(mutex_);
     last_series_by_topic_.clear();
     last_series_fingerprint_by_topic_.clear();
+    // Initialize placeholder series for all configured topics so they appear in the list
+    // immediately, even before any messages are received.
+    for (const auto & entry : configs) {
+      if (entry.first.empty()) {
+        continue;
+      }
+      TrajectorySeriesData placeholder;
+      placeholder.topic = entry.first;
+      placeholder.key =
+        (entry.second == TopicMessageKind::Trajectory) ? "trajectory" : "candidates";
+      placeholder.label =
+        entry.first +
+        (entry.second == TopicMessageKind::Trajectory ? " (Trajectory)" : " (ScoredCandidates)");
+      placeholder.has_received_data = false;
+      last_series_by_topic_[entry.first] = {placeholder};
+    }
   }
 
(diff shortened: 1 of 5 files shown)


common/autoware_trajectory_kinematics_rviz_plugin/src/trajectory_series_manager.cpp
```diff
@@ -170,6 +186,9 @@ void TrajectorySeriesManager::onTrajectory(
   const autoware_planning_msgs::msg::Trajectory::ConstSharedPtr msg, const std::string & topic)
 {
   auto series_vec = trajectoryMsgToSeries(*msg, topic);
+  if (!series_vec.empty()) {
+    series_vec[0].has_received_data = true;
+  }
```
ktro2828, line 191: [Q] Don't we need to set all elements to `true`?

```suggestion
  for (auto & series : series_vec) {
    series.has_received_data = true;
  }
```

Owen-Liuyuxuan: @ktro2828 The code looks like a bug. But as I am checking the behavior on the rviz plugin, I didn't find the problem.. Let me research more.

Owen-Liuyuxuan: @ktro2828 
the `trajectoryMsgToSeries` accept one message and just output a vector of a single TrajectorySeriesData. So it is bad formulation, but not a fatal bug.  

The idea is that ScoredCandidates are returning a list of trajectories. In order to keep both function using similar structure/functions, it forces onTrajectory to also work on a vector of trajectory with size=1

ktro2828: I got it, thank you!

ktro2828 (approved):
LGTM

Owen-Liuyuxuan:
@taikitanaka3 Sorry it seems that we need review from you.

taikitanaka3 (approved):

PR #421, opened 2026-05-02
refactor(planning_data_analyzer): organize EPDMS metric files

bskkimm:
## Description

This PR is the first PR of the full EPDMS patched PR series.

For better organized file tree, the below file tree has been implemented.
This first PR mainly includes reorganization on the file tree and  infrastructure files

### Current file tree (In Progress)
```
│──src
│   └── metrics
│       ├── deviation_metrics.cpp
│       ├── deviation_metrics.hpp
│       ├── metric_types.hpp
│       ├── trajectory_metrics.cpp
│       ├── trajectory_metrics.hpp
│       ├── epdms
│       │   ├── aggregation
│       │   │   ├── epdms_aggregation.cpp
│       │   │   └── epdms_aggregation.hpp
│       │   └── subscores
│       │       ├── drivable_area_compliance.cpp
│       │       ├── drivable_area_compliance.hpp
│       │       ├── driving_direction_compliance.cpp
│       │       ├── driving_direction_compliance.hpp
│       │       ├── ego_progress.cpp
│       │       ├── ego_progress.hpp
│       │       ├── extended_comfort.cpp
│       │       ├── extended_comfort.hpp
│       │       ├── history_comfort.cpp
│       │       ├── history_comfort.hpp
│       │       ├── lane_keeping.cpp
│       │       ├── lane_keeping.hpp
│       │       ├── no_at_fault_collision.cpp
│       │       ├── no_at_fault_collision.hpp
│       │       ├── traffic_light_compliance.cpp
│       │       ├── traffic_light_compliance.hpp
│       │       ├── ttc_within_bound.cpp
│       │       └── ttc_within_bound.hpp
│       └── geometry
│           ├── metric_utils.cpp
│           └── metric_utils.hpp
```

### Final expected file tree
```
│──src
│   └── metrics
│       ├── deviation_metrics.cpp
│       ├── deviation_metrics.hpp
│       ├── metric_types.hpp
│       ├── trajectory_metrics.cpp
│       ├── trajectory_metrics.hpp
│       ├── epdms
│       │   ├── aggregation
│       │   │   ├── epdms_aggregation.cpp
│       │   │   └── epdms_aggregation.hpp
│       │   ├── context
│       │   │   ├── epdms_context.cpp
│       │   │   ├── epdms_context.hpp
│       │   │   └── epdms_types.hpp
│       │   └── subscores
│       │       ├── drivable_area_compliance.cpp
│       │       ├── drivable_area_compliance.hpp
[...]

PR #423, opened 2026-05-06
refactor(planning_data_analyzer): split EPDMS geometry helpers

bskkimm:
### Scope

- PR type: infrastructure.
- Target metric or subscore: EPDMS shared geometry helpers.
- This is PR 2 in the EPDMS patch PR series. PR #421 is already merged.

### What Changed

- Split reusable ego footprint, lanelet query, and comfort-signal helpers out of metric_utils.
- Kept metric_utils as the lightweight common helper entry point used by existing metric code.
- Score semantics changed: no intended score behavior change.

### Full Analyzer Runs

- One full route of Takanawa:
- Successful intended topic generated

## Notes for reviewers

- Follow-up PRs: runtime/topic controls, NC, DAC, DDC, TLC, TTC, LK, HC/EC, EP, aggregation.

### DIFF
--- a/planning/autoware_planning_data_analyzer/src/metrics/geometry/comfort_signal.cpp
+++ b/planning/autoware_planning_data_analyzer/src/metrics/geometry/comfort_signal.cpp
@@ -21,6 +21,7 @@
 #include <algorithm>
 #include <cmath>
 #include <optional>
+#include <utility>
 #include <vector>
 
 namespace autoware::planning_data_analyzer::metrics
@@ -130,8 +131,8 @@ std::vector<double> local_polynomial_filter(
     }
     const auto solution = solve_linear_system(std::move(lhs), std::move(rhs));
     if (solution.has_value()) {
-      output.at(i) = solution->at(static_cast<std::size_t>(derivative_order)) *
-                     factorial(derivative_order);
+      output.at(i) =
+        solution->at(static_cast<std::size_t>(derivative_order)) * factorial(derivative_order);
     }
   }
   return output;
@@ -163,13 +164,14 @@ ComfortSignals compute_comfort_signals(
       std::hypot(input.longitudinal_acceleration_mps2, input.lateral_acceleration_mps2));
   }
 
-  signals.longitudinal_accelerations = local_polynomial_filter(
-    longitudinal_accelerations_raw, times, 8, 2, 0);
-  signals.lateral_accelerations = local_polynomial_filter(lateral_accelerations_raw, times, 8, 2, 0);
+  signals.longitudinal_accelerations =
+    local_polynomial_filter(longitudinal_accelerations_raw, times, 8, 2, 0);
+  signals.lateral_accelerations =
+    local_polynomial_filter(lateral_accelerations_raw, times, 8, 2, 0);
   signals.acceleration_magnitudes =
     local_polynomial_filter(acceleration_magnitudes_raw, times, 8, 2, 0);
-  signals.longitudinal_jerks = local_polynomial_filter(
-    signals.longitudinal_accelerations, times, 15, 2, 1);
+  signals.longitudinal_jerks =
+    local_polynomial_filter(signals.longitudinal_accelerations, times, 15, 2, 1);
   signals.lateral_jerks = local_polynomial_filter(signals.lateral_accelerations, times, 15, 2, 1);
   signals.jerk_magnitudes =
     local_polynomial_filter(signals.acceleration_magnitudes, times, 15, 2, 1);
--- a/planning/autoware_planning_data_analyzer/src/metrics/geometry/comfort_signal.cpp
+++ b/planning/autoware_planning_data_analyzer/src/metrics/geometry/comfort_signal.cpp
@@ -30,6 +30,15 @@ namespace autoware::planning_data_analyzer::metrics
 namespace
 {
 
+constexpr std::size_t kAccelerationFilterWindow = 8U;
+constexpr std::size_t kJerkFilterWindow = 15U;
+constexpr std::size_t kYawAccelerationFilterWindow = 15U;
+constexpr int kFilterPolynomialOrder = 2;
+constexpr int kYawAccelerationPolynomialOrder = 3;
+constexpr int kSmoothingDerivativeOrder = 0;
+constexpr int kFirstDerivativeOrder = 1;
+constexpr int kSecondDerivativeOrder = 2;
+
 double factorial(const int n)
 {
   double value = 1.0;
@@ -164,21 +173,31 @@ ComfortSignals compute_comfort_signals(
       std::hypot(input.longitudinal_acceleration_mps2, input.lateral_acceleration_mps2));
   }
 
-  signals.longitudinal_accelerations =
-    local_polynomial_filter(longitudinal_accelerations_raw, times, 8, 2, 0);
-  signals.lateral_accelerations =
-    local_polynomial_filter(lateral_accelerations_raw, times, 8, 2, 0);
-  signals.acceleration_magnitudes =
-    local_polynomial_filter(acceleration_magnitudes_raw, times, 8, 2, 0);
-  signals.longitudinal_jerks =
-    local_polynomial_filter(signals.longitudinal_accelerations, times, 15, 2, 1);
-  signals.lateral_jerks = local_polynomial_filter(signals.lateral_accelerations, times, 15, 2, 1);
-  signals.jerk_magnitudes =
-    local_polynomial_filter(signals.acceleration_magnitudes, times, 15, 2, 1);
+  signals.longitudinal_accelerations = local_polynomial_filter(
+    longitudinal_accelerations_raw, times, kAccelerationFilterWindow, kFilterPolynomialOrder,
+    kSmoothingDerivativeOrder);
+  signals.lateral_accelerations = local_polynomial_filter(
+    lateral_accelerations_raw, times, kAccelerationFilterWindow, kFilterPolynomialOrder,
+    kSmoothingDerivativeOrder);
+  signals.acceleration_magnitudes = local_polynomial_filter(
+    acceleration_magnitudes_raw, times, kAccelerationFilterWindow, kFilterPolynomialOrder,
+    kSmoothingDerivativeOrder);
+  signals.longitudinal_jerks = local_polynomial_filter(
+    signals.longitudinal_accelerations, times, kJerkFilterWindow, kFilterPolynomialOrder,
+    kFirstDerivativeOrder);
+  signals.lateral_jerks = local_polynomial_filter(
+    signals.lateral_accelerations, times, kJerkFilterWindow, kFilterPolynomialOrder,
+    kFirstDerivativeOrder);
+  signals.jerk_magnitudes = local_polynomial_filter(
+    signals.acceleration_magnitudes, times, kJerkFilterWindow, kFilterPolynomialOrder,
+    kFirstDerivativeOrder);
 
   const auto unwrapped_yaws = unwrap_angles(yaws);
-  signals.yaw_rates = local_polynomial_filter(unwrapped_yaws, times, 15, 2, 1);
-  signals.yaw_accelerations = local_polynomial_filter(unwrapped_yaws, times, 15, 3, 2);
+  signals.yaw_rates = local_polynomial_filter(
+    unwrapped_yaws, times, kJerkFilterWindow, kFilterPolynomialOrder, kFirstDerivativeOrder);
+  signals.yaw_accelerations = local_polynomial_filter(
+    unwrapped_yaws, times, kYawAccelerationFilterWindow, kYawAccelerationPolynomialOrder,
+    kSecondDerivativeOrder);
   return signals;
 }
 
--- a/planning/autoware_planning_data_analyzer/src/metrics/geometry/comfort_signal.cpp
+++ b/planning/autoware_planning_data_analyzer/src/metrics/geometry/comfort_signal.cpp
@@ -1,0 +1,183 @@
+// Copyright 2026 TIER IV, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "comfort_signal.hpp"
+
+#include "metric_utils.hpp"
+
+#include <autoware_utils_math/normalization.hpp>
+
+#include <algorithm>
+#include <cmath>
+#include <optional>
+#include <vector>
+
+namespace autoware::planning_data_analyzer::metrics
+{
+
+namespace
+{
+
+double factorial(const int n)
+{
+  double value = 1.0;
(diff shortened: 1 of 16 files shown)


planning/autoware_planning_data_analyzer/src/metrics/geometry/lanelet_queries.cpp
```diff
@@ -0,0 +1,302 @@
...
+autoware_utils_geometry::LineString2d to_linestring2d(const lanelet::ConstLineString3d & line)
+{
+  autoware_utils_geometry::LineString2d line_2d;
+  for (const auto & point : lanelet::utils::to2D(line)) {
+    line_2d.push_back({point.x(), point.y()});
+  }
+  return line_2d;
+}
+
+bool is_pose_in_intersection(
```
go-sakayori, line 209: This implementation differs behaviorally from the original in `metric_utils.cpp`.
Old: called `find_reference_lanelet` to get the single closest route lanelet, then checked `is_intersection_lanelet` on that one lanelet.
New: delegates to `compute_driving_direction_local_context`, which collects all nearby route-consistent lanelets within a 5 m radius and also checks intersection area polygons with a 0.35 m admissible margin.              

Additionally, the guard condition changed from `isHandlerReady` to `isMapMsgReady`, so there is a window where the old function would have returned a non-null result and the new one returns `std::nullopt`.

The PR description says "no intended score behavior change" — is this change intentional?

planning/autoware_planning_data_analyzer/src/metrics/geometry/ego_footprint.hpp
```diff
@@ -0,0 +1,114 @@
...
+  std::vector<bool> corner_drivable;
+  std::vector<bool> corner_drivable_by_road_border;
+  std::vector<RoadBorderSideTest> road_border_side_tests;
+  lanelet::ConstLanelets road_lanelets;
+  lanelet::ConstLanelets shoulder_lanelets;
+  std::vector<lanelet::ConstPolygon3d> intersection_areas;
+  std::vector<lanelet::ConstPolygon3d> hatched_road_markings;
+  std::vector<lanelet::ConstPolygon3d> parking_lots;
+  std::vector<lanelet::ConstLineString3d> road_border_lines;
+  std::optional<autoware_utils_geometry::Polygon2d> road_border_envelope;
```
go-sakayori: This field is never assigned in `compute_ego_area_evaluation`, so it will always be `std::nullopt`. Every other field in EgoAreaEvaluation is explicitly populated. If this is a placeholder for a future feature, please add a comment to that effect. Otherwise it will confuse callers who check this field expecting valid data.

planning/autoware_planning_data_analyzer/src/metrics/geometry/metric_utils.hpp
```diff
@@ -15,49 +15,22 @@
 #ifndef METRICS__GEOMETRY__METRIC_UTILS_HPP_
 #define METRICS__GEOMETRY__METRIC_UTILS_HPP_
 
-#include <autoware/route_handler/route_handler.hpp>
+#include "ego_footprint.hpp"
```
go-sakayori: Nit: `metric_utils.hpp` now includes `ego_footprint.hpp` and `lanelet_queries.hpp`, which makes the “lightweight common helper” header pull in route handler, lanelet, trajectory, and the new footprint APIs for every existing `metric_utils` consumer.

Since this PR is splitting helpers out, would it be cleaner to keep `metric_utils.hpp` limited to the truly common utilities (`is_vehicle_info_valid`, `get_yaw`, `forward_offset_in_ego_frame`, etc.) and update call sites to include `ego_footprint.hpp` / `lanelet_queries.hpp` directly when they need those APIs?

planning/autoware_planning_data_analyzer/src/metrics/geometry/ego_footprint.hpp
```diff
@@ -0,0 +1,114 @@
...
+{
+  bool multiple_lanes{false};
+  bool non_drivable_area{false};
+  bool in_intersection{false};
+  bool intersection_context_available{false};
+  bool inside_road_border_envelope{false};
+  bool road_border_fallback_used{false};
+};
+
+struct RoadBorderSideTest
```
go-sakayori: The name `RoadBorderSideTest` is confusing — in C++, the "Test" suffix typically implies a unit test fixture or predicate.

go-sakayori (commented):
Many of the functions introduced in this PR (`compute_ego_area_evaluation`, `evaluate_trajectory_footprints`, etc.) have no callers in the current codebase yet. If these are being added ahead of follow-up PRs, that is fine, but it may be worth considering whether to merge the implementation together with the first caller, so unused code does not accumulate on main in the interim.

planning/autoware_planning_data_analyzer/src/metrics/geometry/lanelet_queries.cpp
bskkimm: Agreed. I restored `is_pose_in_intersection()` to the previous closest-reference-route-lanelet behavior, so existing callers keep the old intersection semantics. The broader local-context helper remains separate for follow-up EPDMS subscore PRs where the score semantics will be changed and validated explicitly.

planning/autoware_planning_data_analyzer/src/metrics/geometry/ego_footprint.hpp
bskkimm: Removed the abandoned road-border envelope field and the related flag. The preparatory helper now exposes only the line/segment/probe-based road-border data that is actually represented by the implementation.

planning/autoware_planning_data_analyzer/src/metrics/geometry/metric_utils.hpp
bskkimm: Agreed. I removed the `footprint/lanelet` helper includes from metric_utils.hpp and added direct includes at the call sites that need those APIs, so `metric_utils.hpp` remains a lightweight common helper header.

planning/autoware_planning_data_analyzer/src/metrics/geometry/ego_footprint.hpp
bskkimm: Renamed `RoadBorderSideTest` to `RoadBorderFallbackProbe` so the type reads as runtime fallback/probe data rather than a test fixture or predicate.

bskkimm:
Addressed the review comments. Existing metric behavior is preserved by restoring `is_pose_in_intersection()` to the previous closest-reference-lanelet semantics. The broader EPDMS local-context helper is kept separate and documented as preparatory for later subscore PRs. I also removed the unused road-border envelope field, renamed the road-border runtime record type, and kept `metric_utils.hpp` lightweight by using direct includes at call sites.

planning/autoware_planning_data_analyzer/src/metrics/geometry/lanelet_queries.cpp
```diff
@@ -0,0 +1,303 @@
...
+    polygons.push_back(polygon);
+  }
+}
+
+double normalize_angle(const double angle)
+{
+  return std::atan2(std::sin(angle), std::cos(angle));
+}
+
+autoware_utils_geometry::Polygon2d to_polygon_2d(const lanelet::BasicPolygon2d & polygon)
```
go-sakayori: `to_polygon_2d`, `point_in_lanelet`, and `point_in_polygon` are defined identically in both `ego_footprint.cpp` and `lanelet_queries.cpp` anonymous namespaces. Since they live in unnamed namespaces there's no ODR issue, but the two copies will drift over time. Could we lift them into a small internal header.

planning/autoware_planning_data_analyzer/src/metrics/geometry/comfort_signal.cpp
```diff
@@ -0,0 +1,185 @@
...
+    times.push_back(input.time_s + time_offset_s);
+    yaws.push_back(get_yaw(input.pose.orientation));
+    longitudinal_accelerations_raw.push_back(input.longitudinal_acceleration_mps2);
+    lateral_accelerations_raw.push_back(input.lateral_acceleration_mps2);
+    acceleration_magnitudes_raw.push_back(
+      std::hypot(input.longitudinal_acceleration_mps2, input.lateral_acceleration_mps2));
+  }
+
+  signals.longitudinal_accelerations =
+    local_polynomial_filter(longitudinal_accelerations_raw, times, 8, 2, 0);
```
go-sakayori: Could we promote these values to named constants

planning/autoware_planning_data_analyzer/src/metrics/geometry/lanelet_queries.cpp
```diff
@@ -0,0 +1,303 @@
...
+  const lanelet::ConstPolygon3d & polygon, std::vector<lanelet::ConstPolygon3d> & polygons,
+  std::unordered_set<lanelet::Id> & seen_ids)
+{
+  if (seen_ids.insert(polygon.id()).second) {
+    polygons.push_back(polygon);
+  }
+}
+
+double normalize_angle(const double angle)
+{
```
go-sakayori: nit: `comfort_signal.cpp` already uses `autoware_utils_math::normalize_radian` for the same purpose. The implementations differ slightly (`atan2(sin, cos)`  vs `fmod`-based, with a boundary difference at exactly ±π), but for the  `std::abs(...) <= threshold` comparisons here they're interchangeable. Could we use the shared utility for consistency?

planning/autoware_planning_data_analyzer/src/metrics/geometry/ego_footprint.cpp
```diff
@@ -0,0 +1,743 @@
...
+    max_x = std::max(max_x, point.x());
+    max_y = std::max(max_y, point.y());
+  }
+
+  return lanelet::BoundingBox2d{
+    lanelet::BasicPoint2d{min_x - search_margin_m, min_y - search_margin_m},
+    lanelet::BasicPoint2d{max_x + search_margin_m, max_y + search_margin_m}};
+}
+
+lanelet::ConstLanelets collect_candidate_road_lanelets(
```
go-sakayori, line 98: The `seen_ids.insert(...).second` dedupe-and-push pattern is repeated 4 times in this file: L109,  L143, L167, and L193. `lanelet_queries.cpp` already defines `append_unique_lanelet` and `append_unique_polygon` (L52) for exactly this. Could those be lifted to a shared internal header and reused here?

planning/autoware_planning_data_analyzer/src/metrics/geometry/ego_footprint.cpp
```diff
@@ -0,0 +1,743 @@
...
+  return min_distance_m;
+}
+
+struct ClosestSemanticBoundaryPoint
+{
+  autoware_utils_geometry::Point2d point;
+  double distance_m{std::numeric_limits<double>::infinity()};
+};
+
+double squared_distance(
```
go-sakayori: nit: `autoware_utils_geometry::calc_squared_distance2d` already does this for any Point type with `x()/y()`.

planning/autoware_planning_data_analyzer/src/metrics/geometry/ego_footprint.cpp
```diff
@@ -0,0 +1,743 @@
...
+  std::vector<autoware_utils_geometry::Point2d> vertices;
+  for (const auto & point : footprint.outer()) {
+    if (vertices.empty() || !bg::equals(vertices.front(), point)) {
+      vertices.push_back(point);
+    }
+  }
+  return vertices;
+}
+
+lanelet::BoundingBox2d footprint_bounding_box(
```
go-sakayori, line 80: nit: `footprint_bounding_box` here and `point_bounding_box` in `lanelet_queries.cpp` (L78) share the same "build `BoundingBox2d` with margin" construction.

go-sakayori (changes requested):

planning/autoware_planning_data_analyzer/src/metrics/geometry/lanelet_queries.cpp
bskkimm: Addressed in 6b5a0aee. I moved the duplicated lanelet polygon conversion and point containment helpers into a shared internal lanelet_geometry helper and reused them from both lanelet_queries.cpp and ego_footprint.cpp.

planning/autoware_planning_data_analyzer/src/metrics/geometry/comfort_signal.cpp
bskkimm: Addressed in 6b5a0aee. I replaced the raw filter window/order/derivative arguments with named constants for the acceleration, jerk, yaw-rate, and yaw-acceleration filters.

planning/autoware_planning_data_analyzer/src/metrics/geometry/lanelet_queries.cpp
bskkimm: Addressed in 6b5a0aee. I removed the local normalize_angle helper and now use autoware_utils_math::normalize_radian in lanelet_queries.cpp.

planning/autoware_planning_data_analyzer/src/metrics/geometry/ego_footprint.cpp
bskkimm: Addressed in 6b5a0aee. I lifted the unique-append helpers into the shared lanelet_geometry helper and reused them for lanelets, polygons, and road-border line strings.

planning/autoware_planning_data_analyzer/src/metrics/geometry/ego_footprint.cpp
bskkimm: Addressed in 6b5a0aee. I removed the local squared_distance helper and replaced its uses with autoware_utils_geometry::calc_squared_distance2d. The small Point2d-to-Point conversion keeps this compatible with the local validation underlay while still using the shared distance utility.

planning/autoware_planning_data_analyzer/src/metrics/geometry/ego_footprint.cpp
bskkimm: Addressed in 6b5a0aee. I added a shared make_bounding_box helper and updated both point_bounding_box and footprint_bounding_box to use it for the common BoundingBox2d construction.

planning/autoware_planning_data_analyzer/src/metrics/geometry/ego_footprint.cpp
```diff
@@ -0,0 +1,713 @@
...
+  return min_distance_m;
+}
+
+struct ClosestSemanticBoundaryPoint
+{
+  autoware_utils_geometry::Point2d point;
+  double distance_m{std::numeric_limits<double>::infinity()};
+};
+
+geometry_msgs::msg::Point to_msg_point(const autoware_utils_geometry::Point2d & point)
```
go-sakayori, line 270: nit: `calc_squared_distance2d` (in `autoware_utils_geometry/geometry.hpp`) is a template that should accept `autoware_utils_geometry::Point2d` directly, since `Point2d` is a registered boost geometry type. Could we drop this adapter and pass `Point2d` straight through at the three call sites (L308, L394–395, L413–414)? If there was a compile error that made the adapter necessary, a short comment explaining why would help.

bskkimm: I tried passing Point2d directly first, but the required local pilot-auto.x2 validation underlay still routes `calc_squared_distance2d` through message-style fields and fails to compile with Point2d accessors. I kept the adapter and added a short comment explaining that compatibility reason.

go-sakayori (approved):

PR #424, opened 2026-05-14
feat(planning_data_analyzer): aggregate ADE/FDE only for override scenes

go-sakayori:
## Description

Add a framework to compute ADE/FDE (and the other per-horizon trajectory metrics) restricted to scenes around override events. The analyzer now:

- Subscribes to `/vehicle/status/control_mode` (`autoware_vehicle_msgs/msg/ControlModeReport`) from the input bag (topic name configurable via `control_mode_topic`).
- Detects every `AUTONOMOUS -> MANUAL` transition and opens an override window of `open_loop.override.window_sec` seconds (default `0.5`) starting at the transition timestamp.
- Emits an additional `override_only/*` block in `time_step_based_trajectory_metric.json` containing the same per-horizon statistics as the existing aggregation, computed only over trajectory samples whose timestamp falls inside an override window.
- Reports `override_only/num_windows` so consumers can detect when no overrides were present (statistics are then empty).

The override window logic is implemented as pure free functions in `utils/override_windows.{hpp,cpp}` and covered by unit tests, so it can be exercised without instantiating `OpenLoopEvaluator`.

## How was this PR tested?

- `colcon build --packages-select autoware_planning_data_analyzer` (clean rebuild).
- `colcon test --packages-select autoware_planning_data_analyzer`: 51/51 gtest cases pass, including 10 new `OverrideWindowsTest` cases covering empty/single-event inputs, AUTONOMOUS->MANUAL filtering, multiple transitions, boundary inclusivity, and the RCL_ROS_TIME clock contract.
- Run on a real override scenario bag ([TIER IV internal link](https://evaluation.tier4.jp/evaluation/reports/2ba1da83-a9ee-5728-aef6-693bb3a94923/tests/dc0c9cdb-4988-5c05-b16e-8f6231d50cb7/9a39c00f-3eb3-5a91-a64e-3067d4f9e5d4/ea0f7447-3fcf-5ef6-9b13-4da46b796e69?project_id=x2_dev)). 
- Part of the json file 
```
  "override_only/full/min_ttc/description": "Override-only minimum TTC within full horizon [s]",
  "override_only/full/min_ttc/max": 1.7976931348623157e+308,
  "override_only/full/min_ttc/mean": null,
  "override_only/full/min_ttc/min": 1.7976931348623157e+308,
  "override_only/full/min_ttc/percentile_95": 1.7976931348623157e+308,
  "override_only/full/min_ttc/percentile_99": 1.7976931348623157e+308,
  "override_only/full/min_ttc/std": null,
  "override_only/num_windows": 1
```

## Notes for reviewers
- Override aggregation is keyed off the same horizon tags as the existing aggregation, so the diff in the summary JSON is purely additive (a new `override_only/*` namespace).

## Effects on system behavior
None

### DIFF
--- a/planning/autoware_planning_data_analyzer/CMakeLists.txt
+++ b/planning/autoware_planning_data_analyzer/CMakeLists.txt
@@ -23,6 +23,7 @@ ament_auto_add_library(${PROJECT_NAME} SHARED
   src/metrics/trajectory_metrics.cpp
   src/utils/bag_utils.cpp
   src/utils/json_utils.cpp
+  src/utils/override_windows.cpp
   src/utils/path_utils.cpp
   src/autoware_planning_data_analyzer_node.cpp
 )
@@ -52,6 +53,7 @@ if(BUILD_TESTING)
     test/metrics/test_ttc_within_bound.cpp
     test/metrics/test_no_at_fault_collision.cpp
     test/metrics/test_traffic_light_compliance.cpp
+    test/utils/test_override_windows.cpp
   )
 
   target_include_directories(test_metrics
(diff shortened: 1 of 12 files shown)


xtk8532704 (approved):
LGTM

PR #425, opened 2026-05-15
feat(planning_data_analyzer): add EPDMS runtime topic controls

bskkimm:
### Scope

- PR type: runtime-control / output-topic.
- Target metric or subscore: EPDMS open-loop metric outputs.
- Base branch: `autowarefoundation:main`.
- This is PR 3 in the EPDMS patch PR series.

### What Changed

- Added `open_loop.enabled_metrics`, defaulting to `["all"]`.
- Added `open_loop.debug_topics_enabled`, defaulting to `false`.
- Moved EPDMS score, availability, reason, and aggregate metric topics to `/open_loop/metrics/epdms/*`.
- Kept trajectory diagnostic arrays under `/trajectory/*` and non-EPDMS trajectory metrics under `/open_loop/metrics/<variant>/*`.
- Score semantics changed: no.
- Debug topics changed: default control parameter added; no debug publishers are added in this PR.
- Output topic schema changed: yes, EPDMS metric topics are namespaced.

### Expected Behavior

- Default behavior still evaluates all existing metrics.
- EPDMS output topics are produced under `/open_loop/metrics/epdms/*`.
- Passing a subset such as `["nc", "ttc"]` narrows written metric topics to those EPDMS outputs.

### Full Analyzer Runs

- Takanawa:
  - Input: `/home/beomseokkim2/rosbag/x2_takanawa/input_bag`
  - Output artifact: `/home/beomseokkim2/rosbag/x2_takanawa/run_logs/20260515_215540`
  - Command/script: `fish /tmp/run_takanawa_epdms_defaults_no_debug.fish`
  - Evaluated trajectories: 8655

### Topic Verification

- EPDMS topics produced: `/open_loop/metrics/epdms/*`
- Diagnostic topics produced: `/trajectory/raw/*`
- Debug topics produced: none under `/debug/epdms/*` with default `open_loop.debug_topics_enabled:=false`

## Notes for reviewers

- Review focus: runtime parameter parsing and result-topic registration/writes.
- Follow-up PRs will port individual NAVSIM-faithful EPDMS subscore logic behind these controls.

## Effects on system behavior

- Runtime behavior: users can restrict written metric topics via `open_loop.enabled_metrics`.
- Score behavior: unchanged.
- Debug/output behavior: EPDMS metric topic namespace changes to `/open_loop/metrics/epdms/*`; debug topics are disabled by default.

### DIFF
--- a/planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
+++ b/planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
@@ -345,14 +345,16 @@ void OpenLoopEvaluator::set_enabled_metrics(const std::vector<std::string> & ena
     return;
   }
 
-  bool has_all = false;
-  auto validation_metrics = make_disabled_metrics();
+  std::vector<std::string> normalized_names;
+  normalized_names.reserve(enabled_metric_names.size());
   for (const auto & name : enabled_metric_names) {
     const auto normalized_name = normalize_metric_name(name);
-    enable_metric_name(validation_metrics, normalized_name);
-    has_all = has_all || normalized_name == "all";
+    normalized_names.push_back(normalized_name);
   }
-  if (has_all && enabled_metric_names.size() > 1U) {
+  const bool has_all = std::any_of(
+    normalized_names.begin(), normalized_names.end(),
+    [](const auto & name) { return name == "all"; });
+  if (has_all && normalized_names.size() > 1U) {
     throw std::invalid_argument("open_loop.enabled_metrics 'all' cannot be combined with others");
   }
   if (has_all) {
@@ -361,7 +363,7 @@ void OpenLoopEvaluator::set_enabled_metrics(const std::vector<std::string> & ena
   }
 
   auto enabled_metrics = make_disabled_metrics();
-  for (const auto & name : enabled_metric_names) {
+  for (const auto & name : normalized_names) {
     enable_metric_name(enabled_metrics, name);
   }
   enabled_metrics_ = enabled_metrics;
--- a/planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
+++ b/planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
@@ -37,6 +37,7 @@
 
 #include <algorithm>
 #include <atomic>
+#include <cctype>
 #include <cmath>
 #include <cstdlib>
 #include <exception>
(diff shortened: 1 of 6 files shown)


go-sakayori (commented):
1. **`debug_topics_enabled_` is unused.** The field is declared in the header, stored in the node and evaluator, but never read in any conditional logic in this PR. Fine per the PR description, but it may produce a compiler warning if `-Wunused-private-field` is enabled. Consider a `[[maybe_unused]]` annotation or a note in the header until the follow-up PR adds consumers.

2. **Duplicated guard expression for synthetic EPDMS.** Both `save_metrics_to_bag` and `get_result_topics` independently compute:
```cpp
enabled_metrics_.synthetic_epdms && all_epdms_inputs_enabled(enabled_metrics_)
```
If the guard semantics ever change, there are two places to update. A small private helper would keep them in sync.

3. **`\"all\"` mixed with other names silently wins.** `set_enabled_metrics({"all", "bogus_name"})` returns early on `"all"` and never validates `"bogus_name"`. But `set_enabled_metrics({"nc", "bogus_name"})` throws. Inconsistent — a user who accidentally types `["all", "nc"]` gets no feedback that `"nc"` was redundant. A warning log or a note in the config comment would help.

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
```diff
@@ -1117,109 +1212,152 @@ void OpenLoopEvaluator::save_metrics_to_bag(
...
+  write_array(enabled_metrics_.trajectory_errors, metrics.displacement_errors, metric_topic("fde"));
+  write_array(enabled_metrics_.trajectory_errors, metrics.ahe, metric_topic("ahe"));
+  write_array(enabled_metrics_.trajectory_errors, metrics.heading_errors, metric_topic("fhe"));
+  write_array(
+    enabled_metrics_.trajectory_errors, metrics.lateral_deviations,
+    metric_topic("lateral_deviation"));
+  write_array(
+    enabled_metrics_.trajectory_errors, metrics.longitudinal_deviations,
+    metric_topic("longitudinal_deviation"));
+  write_array(enabled_metrics_.trajectory_errors, metrics.ttc, metric_topic("ttc"));
```
go-sakayori: Is `metrics.ttc` here the same data as `metrics.ttc_values` written in save_trajectory_point_metrics_to_bag_with_variant? If so, should this be gated on time_to_collision_within_bound instead of trajectory_errors? Even if they represent different computations, consolidating them under a single gate would make the intent clearer. Any thoughts?

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
```diff
@@ -263,6 +337,27 @@ static std::string result_summary(bool has_frame, size_t num_points)
 
 // Constructor implementation moved to header file
 
+void OpenLoopEvaluator::set_enabled_metrics(const std::vector<std::string> & enabled_metric_names)
+{
+  if (enabled_metric_names.empty()) {
+    enabled_metrics_ = EnabledOpenLoopMetrics{};
```
go-sakayori, line 344: Should an empty list mean "no metrics"? If the current behavior is intentional, it's worth a comment here, and also noting it in the config YAML.

go-sakayori (changes requested):

bskkimm:
Addressed the review items as below

- Kept `debug_topics_enabled_` intentionally reserved for the follow-up debug-topic PR and documented that in the header/YAML.
- Added `should_write_synthetic_epdms()` and reused it for writing topics and declaring result topics.
- Changed `enabled_metrics` handling so [] remains all-enabled for compatibility, ["all"] means all, and "all" mixed with other names now throws.
- Moved `/open_loop/metrics/ttc` gating to `time_to_collision_within_bound`.
- Added tests for narrowed TTC topics and mixed "all" rejection.

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
bskkimm: Updated. /open_loop/metrics/ttc is now gated by time_to_collision_within_bound in both topic declaration and bag writing, and the narrowed-topic test checks that ttc is present while ADE is absent for {nc, ttc}.

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
bskkimm: Kept the current behavior for compatibility and documented it. Empty enabled_metrics and ["all"] both mean all metrics. I also made "all" mixed with other names throw, with tests for {"all", "nc"} and {"all", "bogus"}.

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
```diff
@@ -61,6 +62,79 @@ namespace autoware::planning_data_analyzer
...
+  metrics.ego_progress = false;
+  metrics.drivable_area_compliance = false;
+  metrics.no_at_fault_collision = false;
+  metrics.driving_direction_compliance = false;
+  metrics.traffic_light_compliance = false;
+  metrics.synthetic_epdms = false;
+  return metrics;
+}
+
+void enable_metric_name(EnabledOpenLoopMetrics & metrics, const std::string & raw_name)
```
go-sakayori, line 90: "all" is not here, so this call throws `std::invalid_argument` when the input is `["all"]`, which is the default config value. Consider either checking for "all" before this call, or adding it as a no-op case inside enable_metric_name.

go-sakayori (changes requested):

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
bskkimm: Fixed. `set_enabled_metrics()` now normalizes the entries first and handles the top-level `"all"` selector before calling `enable_metric_name()`, so `["all"]` keeps the all-enabled default while mixed forms still throw. Added `EXPECT_NO_THROW` coverage for `["all"]` and reran targeted pre-commit plus `test_offline_evaluation`.

go-sakayori (approved):

PR #426, opened 2026-05-18
feat(planning_data_analyzer): migrate NC metric

bskkimm:
## Description

First subscore migration PR in the EPDMS patch series.

This migrates no-at-fault collision (NC) to use logged tracked objects and shared EPDMS helper utilities for object tracks and future-object slicing.

## Logic Change

As-is pipeline:

1. Read one synchronized predicted-object message.
2. Use predicted object states/paths from that single frame.
3. Compare ego trajectory footprints with those predicted objects.
4. Classify NC collision result from that predicted-object overlap.

To-be pipeline:

1. Read the tracked-object timeline from the rosbag.
2. Slice the future tracked-object messages for each evaluated trajectory horizon.
3. Build reusable logged-object tracks by object ID.
4. Interpolate each tracked object to every trajectory sample time.
5. Reuse shared ego footprint polygons for the same trajectory samples.
6. Check ego/object polygon overlap.
7. Classify at-fault NC result using ego motion, object motion, collision side, and lateral-assessment context.

The PR adds more lines because it introduces the reusable tracked-object helper layer, object_tracks, that later EPDMS migrations can reuse instead of duplicating timeline slicing/interpolation logic.

### One Takanawa route run against baseline
- NC exact match, 0 changed records

### DIFF
--- a/planning/autoware_planning_data_analyzer/src/data_types.hpp
+++ b/planning/autoware_planning_data_analyzer/src/data_types.hpp
@@ -87,7 +87,7 @@ struct TopicNames
   std::string control_mode_topic;
   double evaluation_interval_ms = 100.0;
   double sync_tolerance_ms = 100.0;
-  double trajectory_evaluation_horizon_s = 0.0;
+  double trajectory_evaluation_horizon_s = 4.0;
 };
 
 }  // namespace autoware::planning_data_analyzer
--- a/planning/autoware_planning_data_analyzer/src/data_types.hpp
+++ b/planning/autoware_planning_data_analyzer/src/data_types.hpp
@@ -19,6 +19,7 @@
 
 #include <autoware_internal_planning_msgs/msg/candidate_trajectories.hpp>
 #include <autoware_perception_msgs/msg/predicted_objects.hpp>
+#include <autoware_perception_msgs/msg/tracked_objects.hpp>
 #include <autoware_perception_msgs/msg/traffic_light_group_array.hpp>
 #include <autoware_planning_msgs/msg/trajectory.hpp>
 #include <autoware_vehicle_msgs/msg/control_mode_report.hpp>
@@ -29,6 +30,7 @@
 
 #include <memory>
 #include <string>
+#include <vector>
 
 namespace autoware::planning_data_analyzer
 {
@@ -39,10 +41,17 @@ using CandidateTrajectories = autoware_internal_planning_msgs::msg::CandidateTra
 using TFMessage = tf2_msgs::msg::TFMessage;
 using Odometry = nav_msgs::msg::Odometry;
 using PredictedObjects = autoware_perception_msgs::msg::PredictedObjects;
+using TrackedObjects = autoware_perception_msgs::msg::TrackedObjects;
 using TrafficLightGroupArray = autoware_perception_msgs::msg::TrafficLightGroupArray;
 using AccelWithCovarianceStamped = geometry_msgs::msg::AccelWithCovarianceStamped;
 using SteeringReport = autoware_vehicle_msgs::msg::SteeringReport;
 using ControlModeReport = autoware_vehicle_msgs::msg::ControlModeReport;
+
+struct TimedTrackedObjects
+{
+  rclcpp::Time stamp;
+  std::shared_ptr<const TrackedObjects> objects;
+};
 
 // Synchronized data from multiple topics at a specific timestamp
 struct SynchronizedData
@@ -54,6 +63,8 @@ struct SynchronizedData
   std::shared_ptr<AccelWithCovarianceStamped> acceleration;
   std::shared_ptr<SteeringReport> steering_status;
   std::shared_ptr<PredictedObjects> objects;
+  std::shared_ptr<TrackedObjects> tracked_objects;
+  std::vector<TimedTrackedObjects> future_tracked_objects;
   std::shared_ptr<TrafficLightGroupArray> traffic_signals;
   rclcpp::Time timestamp;
   rclcpp::Time bag_timestamp;
@@ -68,6 +79,7 @@ struct TopicNames
   std::string candidate_trajectories_topic;
   std::string gt_trajectory_topic;
   std::string objects_topic;
+  std::string tracked_objects_topic;
   std::string traffic_signals_topic;
   std::string tf_topic;
   std::string acceleration_topic;
@@ -75,6 +87,7 @@ struct TopicNames
   std::string control_mode_topic;
   double evaluation_interval_ms = 100.0;
   double sync_tolerance_ms = 100.0;
+  double trajectory_evaluation_horizon_s = 0.0;
 };
 
 }  // namespace autoware::planning_data_analyzer
--- a/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/no_at_fault_collision.cpp
+++ b/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/no_at_fault_collision.cpp
@@ -24,12 +24,10 @@
 #include <lanelet2_core/geometry/Lanelet.h>
 
 #include <algorithm>
-#include <array>
 #include <cmath>
-#include <cstdint>
 #include <memory>
-#include <set>
 #include <string>
+#include <unordered_set>
 #include <vector>
 
 namespace autoware::planning_data_analyzer::metrics
@@ -44,7 +42,9 @@ using autoware_utils_geometry::Point2d;
 using autoware_utils_geometry::Polygon2d;
 namespace bg = boost::geometry;
 
-constexpr double kStoppedSpeedThreshold = 5.0e-2;
+// Strict NAVSIM-style physical-stop threshold for NC classification; this is intentionally
+// tighter than general perception stopped-vehicle thresholds to keep creeping objects active.
+constexpr double kStoppedVelocityThresholdMps = 5.0e-2;
 
 enum class CollisionType {
   StoppedEgo,
@@ -89,7 +89,7 @@ bool front_bumper_intersects(
 bool is_track_stopped(const InterpolatedLoggedObject & object_state)
 {
   return !is_agent_classification(object_state.classification) ||
-         object_state.speed_mps <= kStoppedSpeedThreshold;
+         object_state.speed_mps <= kStoppedVelocityThresholdMps;
 }
 
 CollisionType classify_collision(
@@ -97,7 +97,7 @@ CollisionType classify_collision(
   const InterpolatedLoggedObject & object_state,
   const autoware::vehicle_info_utils::VehicleInfo & vehicle_info)
 {
-  if (ego_speed(ego_point) <= kStoppedSpeedThreshold) {
+  if (ego_speed(ego_point) <= kStoppedVelocityThresholdMps) {
     return CollisionType::StoppedEgo;
   }
   if (is_track_stopped(object_state)) {
@@ -158,7 +158,7 @@ NoAtFaultCollisionResult calculate_no_at_fault_collision(
     return result;
   }
 
-  std::set<std::array<uint8_t, 16>> collided_object_ids;
+  std::unordered_set<unique_identifier_msgs::msg::UUID, UuidHash> collided_object_ids;
   const auto trajectory_start_time = rclcpp::Time(trajectory.header.stamp);
 
   auto record_at_fault_collision = [&](
--- a/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/no_at_fault_collision.cpp
+++ b/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/no_at_fault_collision.cpp
@@ -24,6 +24,7 @@
 #include <lanelet2_core/geometry/Lanelet.h>
 
 #include <algorithm>
+#include <array>
 #include <cmath>
 #include <cstdint>
 #include <memory>
@@ -54,10 +55,10 @@ enum class CollisionType {
   ActiveLateral,
 };
 
-struct EgoAreaFlags
+struct AtFaultCollision
 {
-  bool multiple_lanes{false};
-  bool non_drivable_area{false};
+  double score{1.0};
+  std::string reason{"available"};
 };
 
 double ego_speed(const autoware_planning_msgs::msg::TrajectoryPoint & point)
@@ -112,49 +113,15 @@ CollisionType classify_collision(
   return CollisionType::ActiveLateral;
 }
 
-bool footprint_intersects_lanelet(
-  const Polygon2d & footprint, const lanelet::ConstLanelet & lanelet)
-{
-  return !bg::disjoint(footprint, lanelet.polygon2d().basicPolygon());
-}
-
-std::optional<EgoAreaFlags> compute_ego_area_flags(
-  const geometry_msgs::msg::Pose & pose, const Polygon2d & ego_polygon,
-  const std::shared_ptr<RouteHandler> & route_handler)
-{
-  if (!route_handler || !route_handler->isHandlerReady()) {
-    return std::nullopt;
-  }
-
-  std::size_t intersecting_route_lanelets = 0;
-  for (const auto & lanelet : route_handler->getRoadLaneletsAtPose(pose)) {
-    if (!route_handler->isRouteLanelet(lanelet)) {
-      continue;
-    }
-    if (footprint_intersects_lanelet(ego_polygon, lanelet)) {
-      ++intersecting_route_lanelets;
-    }
-  }
-
-  EgoAreaFlags flags;
-  flags.non_drivable_area = intersecting_route_lanelets == 0U;
-  flags.multiple_lanes = intersecting_route_lanelets > 1U;
-  return flags;
-}
-
-NoAtFaultCollisionResult make_at_fault_result(
-  const InterpolatedLoggedObject & object, const double time_s, const bool lateral_collision)
+AtFaultCollision make_at_fault_collision(
+  const InterpolatedLoggedObject & object, const bool lateral_collision)
 {
   const bool agent = is_agent_classification(object.classification);
-  NoAtFaultCollisionResult result;
-  result.available = true;
-  result.score = agent ? 0.0 : 0.5;
-  result.infraction_time_s = time_s;
-  result.reason = lateral_collision ? (agent ? "at_fault_lateral_collision_with_agent"
-                                             : "at_fault_lateral_collision_with_non_agent")
-                                    : (agent ? "at_fault_collision_with_agent"
-                                             : "at_fault_collision_with_non_agent");
-  return result;
+  return AtFaultCollision{
+    agent ? 0.0 : 0.5, lateral_collision ? (agent ? "at_fault_lateral_collision_with_agent"
+                                                  : "at_fault_lateral_collision_with_non_agent")
+                                         : (agent ? "at_fault_collision_with_agent"
+                                                  : "at_fault_collision_with_non_agent")};
 }
 
 }  // namespace
@@ -164,6 +131,7 @@ NoAtFaultCollisionResult calculate_no_at_fault_collision(
   const std::vector<TimedTrackedObjects> & future_objects,
   const autoware::vehicle_info_utils::VehicleInfo & vehicle_info,
   const std::shared_ptr<RouteHandler> & route_handler,
+  const std::vector<TrajectoryFootprintEvaluation> * footprint_evaluations,
   const std::vector<LoggedObjectTrack> * object_tracks)
 {
   NoAtFaultCollisionResult result;
@@ -172,7 +140,7 @@ NoAtFaultCollisionResult calculate_no_at_fault_collision(
     result.reason = "unavailable_empty_trajectory";
     return result;
   }
-  if (future_objects.empty() && !object_tracks) {
+  if (future_objects.empty()) {
     result.reason = "unavailable_no_future_objects";
     return result;
   }
@@ -192,14 +160,25 @@ NoAtFaultCollisionResult calculate_no_at_fault_collision(
     return result;
   }
 
-  const auto local_footprint = vehicle_info.createFootprint(0.0);
+  const auto local_evaluations =
+    footprint_evaluations ? std::vector<TrajectoryFootprintEvaluation>{}
+                          : evaluate_trajectory_footprints(trajectory, vehicle_info, route_handler);
+  const auto & evaluations = footprint_evaluations ? *footprint_evaluations : local_evaluations;
+  if (evaluations.size() != trajectory.points.size()) {
+    result.available = false;
+    result.reason = "unavailable_invalid_footprint";
+    result.score = 0.0;
+    return result;
+  }
+
   std::set<std::array<uint8_t, 16>> collided_object_ids;
   const auto trajectory_start_time = rclcpp::Time(trajectory.header.stamp);
 
-  for (const auto & point : trajectory.points) {
+  for (size_t index = 0; index < trajectory.points.size(); ++index) {
+    const auto & point = trajectory.points.at(index);
     const auto query_time = trajectory_start_time + rclcpp::Duration(point.time_from_start);
     const auto query_time_s = rclcpp::Duration(point.time_from_start).seconds();
-    const auto ego_polygon = create_pose_footprint(point.pose, local_footprint);
+    const auto & ego_polygon = evaluations.at(index).ego_polygon;
 
     for (const auto & object_track : tracks) {
       if (
@@ -225,12 +204,18 @@ NoAtFaultCollisionResult calculate_no_at_fault_collision(
       const bool lateral_collision = collision_type == CollisionType::ActiveLateral;
 
       if (front_or_stopped_track) {
-        return make_at_fault_result(*object_state, query_time_s, false);
+        const auto at_fault_collision = make_at_fault_collision(*object_state, false);
+        if (at_fault_collision.score < result.score) {
+          result.score = at_fault_collision.score;
+          result.reason = at_fault_collision.reason;
+          result.infraction_time_s = query_time_s;
+        }
+        continue;
       }
 
       if (lateral_collision) {
-        const auto ego_area_flags = compute_ego_area_flags(point.pose, ego_polygon, route_handler);
-        if (!ego_area_flags.has_value()) {
+        const auto & ego_area_evaluation = evaluations.at(index).ego_area_evaluation;
+        if (!ego_area_evaluation.has_value()) {
           result.available = false;
           result.score = 0.0;
           result.reason = !route_handler
@@ -239,9 +224,17 @@ NoAtFaultCollisionResult calculate_no_at_fault_collision(
           return result;
         }
 
-        if (ego_area_flags->multiple_lanes || ego_area_flags->non_drivable_area) {
-          return make_at_fault_result(*object_state, query_time_s, true);
+        if (
+          ego_area_evaluation->flags.multiple_lanes ||
+          ego_area_evaluation->flags.non_drivable_area) {
+          const auto at_fault_collision = make_at_fault_collision(*object_state, true);
+          if (at_fault_collision.score < result.score) {
+            result.score = at_fault_collision.score;
+            result.reason = at_fault_collision.reason;
+            result.infraction_time_s = query_time_s;
+          }
         }
+        continue;
       }
     }
   }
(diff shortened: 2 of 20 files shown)


planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/no_at_fault_collision.cpp
```diff
@@ -250,43 +153,69 @@ NoAtFaultCollisionResult calculate_no_at_fault_collision(
...
-      if (!object_state.has_value()) {
+      const auto object_state = interpolate_logged_object_state(object_track, query_time);
+      if (!object_state.has_value() || is_unknown_classification(object_state->classification)) {
         continue;
       }
       if (!bg::intersects(ego_polygon, object_state->polygon)) {
         continue;
       }
+      if (object_state->has_valid_object_id) {
+        collided_object_ids.insert(object_state->object_id);
```
go-sakayori: **Logic bug — dedup runs before fault classification.**

`collided_object_ids.insert(...)` here marks the object on *any* polygon intersection, before `classify_collision` is called. For `CollisionType::StoppedEgo` and `CollisionType::RearAgent` the inner loop falls through without scoring, but the object is still recorded as collided. Any later trajectory point where the *same* object becomes an `ActiveFront`/`ActiveLateral` collision is then skipped by the dedup guard at line 186, silently dropping a genuine at-fault collision.

The previous implementation didn't have this problem because it only `return`ed on at-fault types.

Suggested fix: move the insert into the at-fault branches (after the score update), or guard it with `if (front_or_stopped_track || lateral_collision)`.

The single Takanawa run with 0 changed records is exactly the case this would not surface. Please re-validate on a bag with stopped-ego frames adjacent to moving frames, and add a unit test for `StoppedEgo → ActiveFront` on the same object id.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/no_at_fault_collision.cpp
```diff
@@ -250,43 +153,69 @@ NoAtFaultCollisionResult calculate_no_at_fault_collision(
...
+  if (tracks.empty()) {
     return result;
   }
 
-  const auto local_footprint = vehicle_info.createFootprint(0.0);
-  const auto object_caches = build_object_caches(*objects);
+  const auto local_evaluations =
+    footprint_evaluations ? std::vector<TrajectoryFootprintEvaluation>{}
+                          : evaluate_trajectory_footprints(trajectory, vehicle_info, route_handler);
+  const auto & evaluations = footprint_evaluations ? *footprint_evaluations : local_evaluations;
```
go-sakayori: The dual "optional pointer + maybe-built local" pattern for both `object_tracks` and `footprint_evaluations` is awkward, especially since later EPDMS subscores in the series are expected to reuse the same helpers. Picking one model now (always caller-provided, or always built internally with a small wrapper) will save churn in follow-up migrations.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/no_at_fault_collision.cpp
```diff
@@ -182,63 +113,35 @@ CollisionType classify_collision(
...
-bool is_agent_type(const autoware_perception_msgs::msg::PredictedObject & object)
-{
-  using autoware_perception_msgs::msg::ObjectClassification;
-  const auto label = autoware::object_recognition_utils::getHighestProbLabel(object.classification);
-  return autoware::object_recognition_utils::isVehicle(label) ||
-         label == ObjectClassification::PEDESTRIAN || label == ObjectClassification::ANIMAL;
+  const bool agent = is_agent_classification(object.classification);
+  return AtFaultCollision{
+    agent ? 0.0 : 0.5, lateral_collision ? (agent ? "at_fault_lateral_collision_with_agent"
+                                                  : "at_fault_lateral_collision_with_non_agent")
```
go-sakayori: The double-nested ternary here is dense. Two locals (`agent`, plus a `prefix` string for `at_fault_lateral_` vs `at_fault_`) and a flat ternary on the reason would read significantly better without bloating the function.

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
```diff
@@ -145,6 +146,75 @@ std::shared_ptr<SynchronizedData> clone_with_trajectory(
...
+    }
+
+    if (!truncated.points.empty()) {
+      const double previous_t = point_time_s(truncated.points.back());
+      if (previous_t < horizon_s - kTimeEpsilon) {
+        truncated.points.push_back(
+          interpolate_trajectory_point(truncated.points.back(), point, horizon_s));
+      }
+    }
+    return truncated;
```
go-sakayori, line 215: If the trajectory's first point already has `time_from_start > horizon`, this returns an empty trajectory silently: `truncated.points` is empty, so the `!truncated.points.empty()` guard is false, no interpolated point is pushed, and we return. Downstream `num_pts <= 1u` catches it as a failure frame, but the failure mode is masked.

Consider either keeping the first point unconditionally, or logging when truncation produces an empty trajectory.

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.hpp
```diff
@@ -201,6 +201,11 @@ class OpenLoopEvaluator : public BaseEvaluator
 
   void set_debug_topics_enabled(bool enabled) { debug_topics_enabled_ = enabled; }
 
+  void set_trajectory_evaluation_horizon(double horizon_s)
```
go-sakayori, line 205: Inline setter with no validation. The `< 0.0` runtime check lives only in the node constructor, so library callers of `OpenLoopEvaluator` can bypass it. Consider validating (or `assert`-ing) here.

planning/autoware_planning_data_analyzer/src/data_types.hpp
```diff
@@ -68,13 +79,15 @@ struct TopicNames
...
   std::string objects_topic;
+  std::string tracked_objects_topic;
   std::string traffic_signals_topic;
   std::string tf_topic;
   std::string acceleration_topic;
   std::string steering_topic;
   std::string control_mode_topic;
   double evaluation_interval_ms = 100.0;
   double sync_tolerance_ms = 100.0;
+  double trajectory_evaluation_horizon_s = 0.0;
```
go-sakayori: Default here is `0.0` (= full trajectory), but `planning_data_analyzer.param.yaml` sets `4.0`. The same disagreement exists on `OpenLoopEvaluator::trajectory_evaluation_horizon_s_{0.0}`. Anyone constructing `TopicNames`/the evaluator programmatically (tests, library reuse) gets full-trajectory behaviour while production runs at 4 s. Worth aligning the C++ default to 4.0, or at least adding a comment that the param file is authoritative.

planning/autoware_planning_data_analyzer/src/metrics/geometry/metric_utils.cpp
```diff
@@ -44,7 +44,20 @@ double forward_offset_in_ego_frame(
 bool is_agent_behind(
   const geometry_msgs::msg::Pose & ego_pose, const geometry_msgs::msg::Pose & object_pose)
 {
-  return forward_offset_in_ego_frame(ego_pose, object_pose) < 0.0;
+  constexpr double kPi = 3.14159265358979323846;
```
go-sakayori: `M_PI` is used elsewhere in this package (e.g. `object_tracks.cpp`). Introducing a local `kPi` here is inconsistent — please reuse `M_PI`.

Separately: this widens `is_agent_behind` from a half-plane test to a >150° cone, which is a meaningful semantic change for NC classification. A one-line comment naming the source for the `5π/6` threshold (NAVSIM alignment? empirical?) would help future readers.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+double closest_pi_symmetric_yaw(const double reference_yaw, const double yaw)
+{
+  double best_yaw = yaw;
+  double best_error = std::abs(yaw - reference_yaw);
+  for (int multiplier = -2; multiplier <= 2; ++multiplier) {
+    const double candidate_yaw = yaw + static_cast<double>(multiplier) * M_PI;
+    const double candidate_error = std::abs(candidate_yaw - reference_yaw);
+    if (candidate_error < best_error) {
+      best_yaw = candidate_yaw;
```
go-sakayori, line 45: The loop iterates `multiplier ∈ [-2, 2]`, but the π-symmetry of a bounding box only needs `±1·π`. The `±2π` candidates can't change polygon orientation — they're redundant work.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
```diff
@@ -0,0 +1,320 @@
...
+}
+
+bool should_hold_long_object_yaw(
+  const LoggedObjectState & previous_state, const LoggedObjectState & current_state,
+  const double previous_yaw, const double current_yaw)
+{
+  constexpr double kLongObjectLengthThresholdM = 8.0;
+  constexpr double kSlowObjectSpeedThresholdMps = 2.0;
+  constexpr double kSmallDisplacementThresholdM = 0.75;
+  constexpr double kLargeYawJumpThresholdRad = 10.0 * M_PI / 180.0;
```
go-sakayori, line 66: These thresholds (`8 m`, `2 mps`, `0.75 m`, `10°`) are unexplained magic numbers. A one-line comment naming the failure mode they guard against (tracker yaw flips on long stationary objects?) would help future maintainers.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
```diff
@@ -0,0 +1,320 @@
...
+{
+  using autoware_perception_msgs::msg::ObjectClassification;
+  return autoware::object_recognition_utils::getHighestProbLabel(classification) ==
+         ObjectClassification::UNKNOWN;
+}
+
+std::vector<LoggedObjectTrack> build_logged_object_tracks(
+  const std::vector<TimedTrackedObjects> & future_objects)
+{
+  std::map<std::array<uint8_t, 16>, LoggedObjectTrack> keyed_tracks;
```
go-sakayori: `std::map` keyed on `std::array<uint8_t, 16>` gives O(log n) per insert. Since this helper is intended to be reused by other EPDMS subscores against potentially large bags, `std::unordered_map` with a hash on the UUID array would be a cheap win.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
```diff
@@ -0,0 +1,320 @@
...
+  const auto trajectory_start_ns = rclcpp::Time(trajectory.header.stamp).nanoseconds();
+  auto trajectory_horizon_ns =
+    rclcpp::Duration(trajectory.points.back().time_from_start).nanoseconds();
+  if (horizon_s > 0.0) {
+    trajectory_horizon_ns =
+      std::min(trajectory_horizon_ns, rclcpp::Duration::from_seconds(horizon_s).nanoseconds());
+  }
+
+  constexpr rcutils_time_point_value_t kFutureObjectRangeMarginNs =
+    static_cast<rcutils_time_point_value_t>(200'000'000);
```
go-sakayori, line 219: The 200 ms margin is a hidden constant. If the intent is to track `sync_tolerance_ms`, please plumb it in; otherwise a comment explaining the choice would help.

go-sakayori (commented):
Thanks for the migration — the shared `object_tracks` layer is a sensible foundation for the rest of the EPDMS series. Most of my feedback is inline. The one item I'd treat as blocking is the dedup-before-classification in `calculate_no_at_fault_collision` (see inline comment on line 198 of `no_at_fault_collision.cpp`), which can silently drop genuine at-fault collisions and is exactly the kind of regression the single-route validation would not surface.

I'd also like to see tests added for: (a) `StoppedEgo → ActiveFront` on the same object id (covers the dedup bug), and (b) `truncate_trajectory_by_horizon` corner cases (first point beyond horizon, single-point trajectory, exact horizon).

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/no_at_fault_collision.cpp
bskkimm: Fixed. `collided_object_ids` is now updated only when the overlap is classified as an at-fault collision, so ignored stopped-ego/rear cases no longer mask later at-fault collisions with the same object. Added a regression test for the stopped-ego -> later at-fault same-ID case.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/no_at_fault_collision.cpp
bskkimm: Updated. The production NC path now builds shared `LoggedObjectTrack` and footprint evaluations in the caller and passes them to the scorer directly, removing the optional pointer/maybe-local pattern from the main scorer.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/no_at_fault_collision.cpp
bskkimm: Fixed. Split the nested ternary into explicit `score`, `collision_scope`, and `object_kind` locals before constructing the reason string.

planning/autoware_planning_data_analyzer/src/data_types.hpp
bskkimm: Fixed. Aligned the C++ defaults to the YAML default of 4.0 seconds.

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.hpp
bskkimm: Fixed. Added non-negative validation to the setter, matching the node-side parameter validation.

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
bskkimm: Fixed. `truncate_trajectory_by_horizon()` now keeps the first point if the first sample is already beyond the horizon, so it no longer returns an empty trajectory for that edge case.

planning/autoware_planning_data_analyzer/src/metrics/geometry/metric_utils.cpp
bskkimm: Fixed. Reused `M_PI` and added a comment explaining the 150-degree behind-object threshold.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
bskkimm: Fixed. Added a comment explaining these thresholds as a guard against tracker yaw flips for long slow/stationary objects.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
bskkimm: Fixed. Reduced the yaw candidates to the necessary `[-1, 1]` multiplier range for pi-symmetric bounding boxes.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
bskkimm: Fixed. Switched valid object track grouping to `std::unordered_map` with a UUID-array hash, then sorted the output tracks by UUID to keep deterministic iteration.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
bskkimm: Fixed. Added a comment clarifying that the 200 ms range extension is interpolation boundary slack, not message sync tolerance.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
```diff
@@ -0,0 +1,341 @@
...
+
+struct ObjectIdHash
+{
+  std::size_t operator()(const std::array<uint8_t, 16> & object_id) const
+  {
+    std::size_t seed = 0U;
+    for (const auto byte : object_id) {
+      seed ^= std::hash<uint8_t>{}(byte) + 0x9e3779b9U + (seed << 6U) + (seed >> 2U);
+    }
+    return seed;
```
go-sakayori: This is byte-for-byte the same algorithm as `UUIDHash` in `autoware/universe/perception/autoware_multi_object_tracker/.../object_model/types.hpp` (same `0x9e3779b9` seed, same shift pattern). Autoware also already has the canonical pattern via `autoware::universe_utils::toBoostUUID` + `boost::hash<boost::uuids::uuid>` from `<boost/uuid/uuid_hash.hpp>` (used in `behavior_path_planner_common` and `autoware_universe_utils/ros/uuid_helper.hpp`).

Suggest dropping `ObjectIdHash` *and* `object_id_key` below, and changing the map/set key from `std::array<uint8_t, 16>` to `unique_identifier_msgs::msg::UUID` so the canonical hash applies directly:

```cpp
#include <autoware/universe_utils/ros/uuid_helper.hpp>
#include <boost/uuid/uuid_hash.hpp>

struct UuidHash {
  std::size_t operator()(const unique_identifier_msgs::msg::UUID & u) const {
    return boost::hash<boost::uuids::uuid>{}(autoware::universe_utils::toBoostUUID(u));
  }
};
```

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
```diff
@@ -0,0 +1,341 @@
...
+  }
+};
+
+double closest_pi_symmetric_yaw(const double reference_yaw, const double yaw)
+{
+  double best_yaw = yaw;
+  double best_error = std::abs(yaw - reference_yaw);
+  for (int multiplier = -1; multiplier <= 1; ++multiplier) {
```
go-sakayori: Iterating over a tiny set of literals — `for (const int multiplier : {-1, 0, 1})` reads better and removes the boundary expression.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
```diff
@@ -0,0 +1,341 @@
...
+      best_yaw = candidate_yaw;
+      best_error = candidate_error;
+    }
+  }
+  return best_yaw;
+}
+
+double normalize_angle(const double angle)
+{
+  return std::atan2(std::sin(angle), std::cos(angle));
```
go-sakayori: `autoware_utils_math::normalize_radian` already does this (in `autoware_utils_math/include/autoware_utils_math/normalization.hpp`), and `autoware_utils_geometry/geometry.hpp` re-exports it as `using autoware_utils_math::normalize_radian;` — which this TU already includes. The existing helper uses `fmod` (one op) vs the local `atan2(sin, cos)` (two trig calls). Suggest replacing this local helper with a direct call to `autoware_utils_math::normalize_radian`.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
```diff
@@ -0,0 +1,341 @@
...
+
+double planar_speed_mps(const geometry_msgs::msg::Twist & twist)
+{
+  return std::hypot(twist.linear.x, twist.linear.y);
+}
+
+double planar_displacement_m(
+  const geometry_msgs::msg::Pose & lhs, const geometry_msgs::msg::Pose & rhs)
+{
+  return autoware_utils_geometry::calc_distance2d(lhs.position, rhs.position);
```
go-sakayori: Trivial wrapper around `autoware_utils_geometry::calc_distance2d`, which is already used elsewhere in the file. Suggest removing this helper and calling `calc_distance2d` at the single use site below.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
```diff
@@ -0,0 +1,341 @@
...
+
+}  // namespace
+
+bool has_valid_object_id(const unique_identifier_msgs::msg::UUID & object_id)
+{
+  return std::any_of(
+    object_id.uuid.begin(), object_id.uuid.end(), [](const auto byte) { return byte != 0U; });
+}
+
+std::array<uint8_t, 16> object_id_key(const unique_identifier_msgs::msg::UUID & object_id)
```
go-sakayori: Only needed because the map key is `std::array<uint8_t, 16>` instead of `unique_identifier_msgs::msg::UUID`. If the hash recommendation above is taken, this helper goes away — `object.object_id` can be used as the key directly.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
```diff
@@ -0,0 +1,341 @@
...
+{
+  return object_id.uuid;
+}
+
+bool is_agent_classification(
```
go-sakayori, line 130: `object_recognition_utils` already provides `isVehicle`, `isCarLikeVehicle`, `isLargeVehicle` with overloads for `label`, `ObjectClassification`, and `std::vector<ObjectClassification>` — but no `isAgent`. Since later EPDMS subscores in this series will likely need the same vehicle ∪ pedestrian ∪ animal predicate, consider upstreaming this as `object_recognition_utils::isAgent` instead of carrying it here. `is_unknown_classification` could go alongside it (or just be inlined — it's a one-line wrapper).

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
```diff
@@ -0,0 +1,341 @@
...
+  const auto first = std::lower_bound(
+    object_timeline.begin(), object_timeline.end(), trajectory_start_ns,
+    [](const TimedTrackedObjects & timed_objects, const rcutils_time_point_value_t stamp_ns) {
+      return timed_objects.stamp.nanoseconds() < stamp_ns;
+    });
+
+  for (auto itr = first; itr != object_timeline.end(); ++itr) {
+    if (itr->stamp.nanoseconds() > range_end_ns) {
+      break;
+    }
```
go-sakayori: Since the timeline is sorted by `stamp` (guaranteed by `process_bag_common`), this `break`-on-predicate loop can be replaced with a binary search + range copy:

```cpp
const auto last = std::upper_bound(
  first, object_timeline.end(), range_end_ns,
  [](rcutils_time_point_value_t stamp_ns, const TimedTrackedObjects & timed) {
    return stamp_ns < timed.stamp.nanoseconds();
  });
future_objects.assign(first, last);
```

O(log n) instead of O(n) to find the end, removes the explicit loop, and makes the sorted-timeline assumption self-documenting.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/no_at_fault_collision.cpp
```diff
@@ -163,14 +86,21 @@ bool front_bumper_intersects(
   return bg::intersects(front_bumper, object_polygon);
 }
 
+bool is_track_stopped(const InterpolatedLoggedObject & object_state)
+{
+  return !is_agent_classification(object_state.classification) ||
+         object_state.speed_mps <= kStoppedSpeedThreshold;
```
go-sakayori: Two suggestions on this constant:

1. **Naming**: autoware uses `stopped_velocity_threshold` consistently (`autoware_perception_online_evaluator`, `autoware_motion_velocity_run_out_module`, `autoware_motion_velocity_road_user_stop_module`). Suggest renaming to `kStoppedVelocityThreshold` (or `kStoppedVelocityThresholdMps`) to match the ecosystem vocabulary.

2. **Value reference**: `0.05` m/s is much tighter than `autoware_perception_online_evaluator`'s `1.0` m/s default — that's defensible here because NC needs to distinguish "physically stationary" from "creeping," but a one-line comment naming the source (NAVSIM spec? empirical?) would help reviewers who instinctively expect a YAML knob like the other modules use.

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
```diff
@@ -145,6 +146,78 @@ std::shared_ptr<SynchronizedData> clone_with_trajectory(
...
+          interpolate_trajectory_point(truncated.points.back(), point, horizon_s));
+      }
+    }
+    if (truncated.points.empty()) {
+      truncated.points.push_back(point);
+    }
+    return truncated;
+  }
+
+  return truncated;
```
go-sakayori, line 218: `truncate_trajectory_by_horizon` corner cases still don't have unit tests: (a) first point already past horizon, (b) single-point trajectory, (c) point exactly at horizon. The behaviour is now defensive enough that I wouldn't block on this, but worth adding before the next subscore in the series starts depending on truncation behaviour.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
bskkimm: Updated. Replaced the custom UUID array hash/key with UUID message keys and a hash based on autoware_utils::to_boost_uuid + boost::hash<boost::uuids::uuid>, and removed object_id_key.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
bskkimm: Fixed. Removed the local normalization helper and now use autoware_utils_math::normalize_radian.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
bskkimm: Fixed. Changed the multiplier loop to iterate over {-1, 0, 1} directly.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
bskkimm: Fixed. Removed the trivial displacement wrapper and call autoware_utils_geometry::calc_distance2d directly.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
bskkimm: Kept this local for this PR to avoid expanding scope into object_recognition_utils; I will promote it to a shared utility in a follow-up if later EPDMS subscores need the same predicate.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
bskkimm: Fixed together with the UUID-key update. object_id_key is removed and object.object_id is used directly as the key.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/no_at_fault_collision.cpp
bskkimm: Fixed. Renamed to kStoppedVelocityThresholdMps and added a comment explaining the strict 0.05 m/s NC threshold.

planning/autoware_planning_data_analyzer/src/metrics/geometry/object_tracks.cpp
bskkimm: Fixed. Replaced the loop with upper_bound and assign(first, last) using the sorted timeline assumption.

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
bskkimm: Fixed. Added truncation coverage for first point beyond horizon, single-point trajectory, and exact-horizon point.

go-sakayori (approved):

PR #427, opened 2026-05-19
feat(planning_data_analyzer): add NAVSIM-style DAC metric

bskkimm:
### Scope

- PR type: subscore.
- Target metric: EPDMS drivable area compliance (DAC).
- Base branch: `upstream/main`.
- This is the DAC PR in the gradual EPDMS patch series after NC.

### What Changed

- Migrates DAC from the old route-lanelet hull check to the shared ego-footprint semantic drivable-area evaluation.
- Uses existing `TrajectoryFootprintEvaluation` output instead of recomputing ego footprints in the DAC scorer.
- Adds DAC debug summary material and unit coverage for road lane, shoulder, hatched road marking, road-border fallback, and outside-area failure.
- Score semantics changed: yes, DAC now follows the NAVSIM-style corner-in-drivable-area check.
- Debug topics changed: no new topic contract in this PR; debug data remains behind `open_loop.debug_topics_enabled`.
- Output topic or JSON keys changed: no.
- DAC reason values changed: yes, e.g.
  `unavailable_footprint_evaluation_size_mismatch`,
  `non_compliant_corner_outside_drivable_area`, and route-handler unavailable reasons.
- Duplication audit: performed; repeated footprint/map queries are handled through shared ego-footprint evaluation; remaining repetition is limited to debug field serialization.

### Logic Change

As-is pipeline:

1. Build ego footprints along the trajectory.
2. Collect route-relevant drivable lanelets.
3. Fuse candidate lanelet polygons.
4. Pass only if each full footprint stays within the fused lanelet area.

To-be pipeline:

1. Reuse per-sample `TrajectoryFootprintEvaluation` prepared by the shared EPDMS geometry context.
2. For each ego footprint corner, check semantic drivable area: road lanelets, road shoulders, intersection areas, hatched road markings, parking lots, and the existing conservative road-border fallback evidence.
3. Fail DAC when any evaluated sample has a non-drivable corner.
4. Report the first failing sample and candidate-area counts for debugging.

Line-count note:

- Additions are larger than deletions because the PR exposes DAC-specific debug details and tests while reusing shared footprint/context infrastructure added by earlier PRs.

### Local Checks

- Build:
  - Command: `CCACHE_DISABLE=1 colcon build --symlink-install --packages-select autoware_planning_data_analyzer --cmake-args -DCMAKE_BUILD_TYPE=RelWithDebInfo`
  - Result: passed with temporary validation-only `pilot-auto.x2` lanelet API compatibility patch, restored before push.
- Unit tests:
  - Command: `build/autoware_planning_data_analyzer/test_metrics`
  - Result: 61 passed.
  - Command: `build/autoware_planning_data_analyzer/test_offline_evaluation`
  - Result: 20 passed.
- Pre-commit:
  - Command: `pre-commit run --all-files`
  - Result: passed.

### Full Analyzer Runs

- Takanawa:
  - Input: `/home/beomseokkim2/rosbag/x2_takanawa/input_bag`
  - Output artifact: `/home/beomseokkim2/rosbag/x2_takanawa/run_logs/20260519_165136`
  - Command/script: `/tmp/run_takanawa_epdms_defaults_no_debug.fish`
  - Inference time: `192.57s`
  - Analyzer time: `193.13s`
- Odaiba:
  - Not run; Takanawa-only is the required validation bag for this PR series.

### Metric Comparison

- Baseline artifact: `/home/beomseokkim2/rosbag/x2_takanawa/run_logs/20260506_155345/eval_json/time_step_based_trajectory_detailed_result.json`
- PR artifact: `/home/beomseokkim2/rosbag/x2_takanawa/run_logs/20260519_165136/eval_json/time_step_based_trajectory_detailed_result.json`
- Total evaluated trajectories: `8695`.
- DAC comparison: exact match.
- DAC availability: `8695` available, `0` unavailable.
- DAC non-1/failure count: `396` baseline, `396` PR.
- DAC reason counts: `compliant: 8299`, `non_compliant_corner_outside_drivable_area: 396`.
- Representative changed timestamps: N/A for DAC; changed timestamp count is `0`.
- NC comparison: shows the accepted PR #426 semantic delta (`74 -> 87` non-1) from the already-merged NC de-dup fix.
- Other EPDMS subscores are not compared as regression targets in this DAC PR because they are not yet migrated in the upstream PR series.

### Topic Verification

- Metric topics produced: EPDMS metric topics remain under `/open_loop/metrics/epdms/*`.
- Debug topics produced: disabled by default in this validation run.
- Existing ported topics still produced: yes.

## Notes for reviewers

- Review focus: DAC semantic area sources, use of shared footprint evaluations, and DAC-specific debug/test coverage.
- Known limitation: road-border evidence is the conservative fallback already present in the shared ego-footprint context; this PR only activates that existing context for DAC.
- Follow-up PRs: DDC, TLC, TTC, LK, HC/EC, EP, aggregation/human-filter.

## Effects on system behavior

- Runtime behavior: DAC uses shared per-sample footprint/map evaluation rather than route-lanelet polygon fusion.
- Score behavior: DAC score semantics change intentionally.
- Debug/output behavior: no topic or JSON-key contract change; debug output remains opt-in.
- Reason behavior: DAC reason values changed intentionally to describe the new semantic-corner
  failure and unavailable cases.
- Backward compatibility: existing metric topic names and JSON keys are preserved.

### DIFF
--- a/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/drivable_area_compliance.cpp
+++ b/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/drivable_area_compliance.cpp
@@ -54,7 +54,7 @@ std::vector<geometry_msgs::msg::Point> lanelet_polygon_to_points(
   const auto polygon = lanelet.polygon2d().basicPolygon();
   points.reserve(polygon.size());
   for (const auto & point : polygon) {
-    points.push_back(autoware::lanelet2_utils::to_ros(point, z));
+    points.push_back(autoware::experimental::lanelet2_utils::to_ros(point, z));
   }
   return points;
 }
@@ -65,7 +65,8 @@ std::vector<geometry_msgs::msg::Point> map_polygon_to_points(
   std::vector<geometry_msgs::msg::Point> points;
   points.reserve(polygon.size());
   for (const auto & point : polygon) {
-    points.push_back(autoware::lanelet2_utils::to_ros(lanelet::utils::to2D(point), z));
+    points.push_back(
+      autoware::experimental::lanelet2_utils::to_ros(lanelet::utils::to2D(point), z));
   }
   return points;
 }
@@ -76,7 +77,8 @@ std::vector<geometry_msgs::msg::Point> line_string_to_points(
   std::vector<geometry_msgs::msg::Point> points;
   points.reserve(line_string.size());
   for (const auto & point : line_string) {
-    points.push_back(autoware::lanelet2_utils::to_ros(lanelet::utils::to2D(point), z));
+    points.push_back(
+      autoware::experimental::lanelet2_utils::to_ros(lanelet::utils::to2D(point), z));
   }
   return points;
 }
--- a/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/drivable_area_compliance.cpp
+++ b/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/drivable_area_compliance.cpp
@@ -14,9 +14,9 @@
 
 #include "drivable_area_compliance.hpp"
 
-#include "../../geometry/lanelet_geometry.hpp"
-#include "../../geometry/metric_utils.hpp"
+#include "metrics/geometry/metric_utils.hpp"
 
+#include <autoware/lanelet2_utils/conversion.hpp>
 #include <autoware_utils_geometry/boost_geometry.hpp>
 #include <rclcpp/duration.hpp>
 
@@ -35,23 +35,13 @@ namespace autoware::planning_data_analyzer::metrics
 namespace
 {
 
-geometry_msgs::msg::Point to_msg_point(
-  const autoware_utils_geometry::Point2d & point, const double z = 0.0)
-{
-  geometry_msgs::msg::Point msg;
-  msg.x = point.x();
-  msg.y = point.y();
-  msg.z = z;
-  return msg;
-}
-
 std::vector<geometry_msgs::msg::Point> polygon_to_points(
   const autoware_utils_geometry::Polygon2d & polygon, const double z)
 {
   std::vector<geometry_msgs::msg::Point> points;
   points.reserve(polygon.outer().size());
   for (const auto & point : polygon.outer()) {
-    points.push_back(to_msg_point(point, z));
+    points.push_back(autoware_utils_geometry::create_point(point.x(), point.y(), z));
   }
   return points;
 }
@@ -59,21 +49,33 @@ std::vector<geometry_msgs::msg::Point> polygon_to_points(
 std::vector<geometry_msgs::msg::Point> lanelet_polygon_to_points(
   const lanelet::ConstLanelet & lanelet, const double z)
 {
-  return polygon_to_points(to_polygon_2d(lanelet.polygon2d().basicPolygon()), z);
+  std::vector<geometry_msgs::msg::Point> points;
+  const auto polygon = lanelet.polygon2d().basicPolygon();
+  points.reserve(polygon.size());
+  for (const auto & point : polygon) {
+    points.push_back(autoware::lanelet2_utils::to_ros(point, z));
+  }
+  return points;
 }
 
 std::vector<geometry_msgs::msg::Point> map_polygon_to_points(
   const lanelet::ConstPolygon3d & polygon, const double z)
 {
-  return polygon_to_points(to_polygon_2d(lanelet::utils::to2D(polygon).basicPolygon()), z);
+  std::vector<geometry_msgs::msg::Point> points;
+  points.reserve(polygon.size());
+  for (const auto & point : polygon) {
+    points.push_back(autoware::lanelet2_utils::to_ros(lanelet::utils::to2D(point), z));
+  }
+  return points;
 }
 
 std::vector<geometry_msgs::msg::Point> line_string_to_points(
   const lanelet::ConstLineString3d & line_string, const double z)
 {
   std::vector<geometry_msgs::msg::Point> points;
-  for (const auto & point : lanelet::utils::to2D(line_string)) {
-    points.push_back(to_msg_point({point.x(), point.y()}, z));
+  points.reserve(line_string.size());
+  for (const auto & point : line_string) {
+    points.push_back(autoware::lanelet2_utils::to_ros(lanelet::utils::to2D(point), z));
   }
   return points;
 }
@@ -98,20 +100,24 @@ void append_first_failure_debug_info(
   debug_info.corner_count_inside =
     std::count(area.corner_drivable.begin(), area.corner_drivable.end(), true);
   if (!area.footprint_points.empty()) {
-    debug_info.label_anchor = to_msg_point(area.footprint_points.front(), z);
+    const auto & point = area.footprint_points.front();
+    debug_info.label_anchor = autoware_utils_geometry::create_point(point.x(), point.y(), z);
   }
 
   for (std::size_t corner_index = 0; corner_index < area.footprint_points.size(); ++corner_index) {
     if (!area.corner_drivable.at(corner_index)) {
       debug_info.failing_corner_indices.push_back(corner_index);
+      const auto & point = area.footprint_points.at(corner_index);
       debug_info.failing_corners.push_back(
-        {time_s, corner_index, to_msg_point(area.footprint_points.at(corner_index), z)});
+        {time_s, corner_index, autoware_utils_geometry::create_point(point.x(), point.y(), z)});
     }
     if (
       corner_index < area.corner_drivable_by_road_border.size() &&
       area.corner_drivable_by_road_border.at(corner_index)) {
+      const auto & point = area.footprint_points.at(corner_index);
       debug_info.road_border_fallback_corners.push_back(
-        {time_s, corner_index, to_msg_point(area.footprint_points.at(corner_index), z + 0.01)});
+        {time_s, corner_index,
+         autoware_utils_geometry::create_point(point.x(), point.y(), z + 0.01)});
     }
   }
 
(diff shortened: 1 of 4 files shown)


planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/drivable_area_compliance.cpp
```diff
@@ -14,17 +14,19 @@
 
 #include "drivable_area_compliance.hpp"
 
-#include "metrics/geometry/metric_utils.hpp"
+#include "../../geometry/lanelet_geometry.hpp"
+#include "../../geometry/metric_utils.hpp"
```
go-sakayori: The `../../geometry/...` form breaks the convention used by every other file in this directory — `ego_progress.cpp`, `traffic_light_compliance.cpp`, `ttc_within_bound.cpp`, `no_at_fault_collision.cpp`, etc. all use the path-prefixed style that resolves via `target_include_directories`. Google's C++ style guide also explicitly says includes should use the full project-source-relative path, not `.`/`..`. Please switch to:

```cpp
#include "metrics/geometry/lanelet_geometry.hpp"
#include "metrics/geometry/metric_utils.hpp"
```

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/drivable_area_compliance.hpp
```diff
@@ -15,28 +15,90 @@
 #ifndef METRICS__EPDMS__SUBSCORES__DRIVABLE_AREA_COMPLIANCE_HPP_
 #define METRICS__EPDMS__SUBSCORES__DRIVABLE_AREA_COMPLIANCE_HPP_
 
+#include "../../geometry/ego_footprint.hpp"
```
go-sakayori: Same as the cpp side: please use `#include "metrics/geometry/ego_footprint.hpp"` to match the convention used by sibling subscore headers and Google C++ style (no `..` in include paths).

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/drivable_area_compliance.cpp
```diff
@@ -33,151 +35,222 @@ namespace autoware::planning_data_analyzer::metrics
...
-      autoware_utils_geometry::transform_vector(
-        base_footprint, autoware_utils_geometry::pose2transform(point.pose)));
-  }
-
-  return vehicle_footprints;
+  geometry_msgs::msg::Point msg;
+  msg.x = point.x();
+  msg.y = point.y();
+  msg.z = z;
+  return msg;
```
go-sakayori: `to_msg_point` duplicates two existing helpers:

- `autoware_utils_geometry::create_point(x, y, z)` in `autoware_utils_geometry/geometry.hpp:295` — identical body.
- The same-named TU-local helper in `src/metrics/geometry/ego_footprint.cpp:270` (sans the `z` parameter).

Please drop this helper and call `autoware_utils_geometry::create_point(point.x(), point.y(), z)` directly at the call sites. If the `Point2d`-plus-z shape is genuinely needed in more than one TU, hoist a single utility into `metric_utils`/`lanelet_geometry` rather than carrying per-file copies.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/drivable_area_compliance.cpp
```diff
@@ -33,151 +35,222 @@ namespace autoware::planning_data_analyzer::metrics
...
-  for (const auto & point : polygon) {
-    converted.outer().push_back(Point2d(point.x(), point.y()));
+  std::vector<geometry_msgs::msg::Point> points;
+  for (const auto & point : lanelet::utils::to2D(line_string)) {
+    points.push_back(to_msg_point({point.x(), point.y()}, z));
   }
-  boost::geometry::correct(converted);
-  return converted;
+  return points;
 }
```
go-sakayori, line 81: These three helpers (`lanelet_polygon_to_points`, `map_polygon_to_points`, `line_string_to_points`) reinvent `autoware::lanelet2_utils::to_ros` in `autoware_lanelet2_utils/conversion.hpp:59-62`, which already provides:

```cpp
geometry_msgs::msg::Point to_ros(const lanelet::BasicPoint3d & src);
geometry_msgs::msg::Point to_ros(const lanelet::ConstPoint3d & src);
geometry_msgs::msg::Point to_ros(const lanelet::BasicPoint2d & src, const double & z = 0.0);
geometry_msgs::msg::Point to_ros(const lanelet::ConstPoint2d & src, const double & z = 0.0);
```

The current approach also double-converts via `to_polygon_2d(...)` before iterating, which is unnecessary. Suggested simplification:

```cpp
std::vector<geometry_msgs::msg::Point> lanelet_polygon_to_points(
  const lanelet::ConstLanelet & lanelet, const double z)
{
  std::vector<geometry_msgs::msg::Point> points;
  for (const auto & p : lanelet.polygon2d().basicPolygon()) {
    points.push_back(autoware::lanelet2_utils::to_ros(p, z));
  }
  return points;
}
```

This also matches the migration note on the deprecated `autoware_lanelet2_extension::utility::toGeomMsgPt` overloads ("please use `autoware::lanelet2_utils::to_ros`").

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/drivable_area_compliance.cpp
```diff
@@ -33,151 +35,222 @@ namespace autoware::planning_data_analyzer::metrics
...
-  if (!fused_lanelet_polygons) {
-    result.reason = "unavailable_invalid_drivable_area_polygon";
+  const auto local_evaluations =
+    footprint_evaluations != nullptr
+      ? std::vector<TrajectoryFootprintEvaluation>{}
+      : evaluate_trajectory_footprints(trajectory, vehicle_info, route_handler);
+  const auto & evaluations =
+    footprint_evaluations != nullptr ? *footprint_evaluations : local_evaluations;
+  if (evaluations.size() != trajectory.points.size()) {
+    result.reason = "unavailable_invalid_footprint";
```
go-sakayori: When `footprint_evaluations != nullptr` and the caller passes a mismatched-size vector, this branch fires — but the reason `unavailable_invalid_footprint` is misleading because no footprint is built on this path. Consider renaming to something like `unavailable_footprint_evaluation_size_mismatch` so the reason string describes the actual failure mode.

Separately: dropping `unavailable_no_drivable_lanelets` and renaming `non_compliant_footprint_outside_drivable_area` → `non_compliant_corner_outside_drivable_area` is a `reason`-field schema change. The PR description says the JSON schema didn't change, but reason values are part of the per-trajectory output and have changed — worth calling out for downstream dashboards/queries that filter on these strings.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/drivable_area_compliance.cpp
```diff
@@ -33,151 +35,222 @@ namespace autoware::planning_data_analyzer::metrics
...
     return result;
   }
 
+  fill_debug_info(result.debug_info, trajectory, evaluations);
```
go-sakayori: `fill_debug_info(...)` runs before the loop below that may set `available = false` and `reason = "unavailable_drivable_area_query_failed"`. On that early return, `result.debug_info` may already contain ego footprints — and possibly a `first_failure_time_s`/`failing_corners` entry — from earlier samples, while the result is unavailable. Consider either moving `fill_debug_info` after the availability loop, or clearing `result.debug_info` on the unavailable-return path, so consumers can rely on "unavailable ⇒ no debug payload".

planning/autoware_planning_data_analyzer/test/metrics/test_drivable_area_compliance.cpp
```diff
@@ -65,32 +147,77 @@ autoware::vehicle_info_utils::VehicleInfo make_vehicle_info()
...
   EXPECT_DOUBLE_EQ(result.score, 0.0);
-  EXPECT_EQ(result.reason, "non_compliant_footprint_outside_drivable_area");
+  EXPECT_EQ(result.reason, "non_compliant_corner_outside_drivable_area");
+  EXPECT_TRUE(std::isfinite(result.debug_info.first_failure_time_s));
+  EXPECT_FALSE(result.debug_info.failing_corners.empty());
+  EXPECT_FALSE(result.debug_info.ego_horizon_footprints.empty());
+  EXPECT_FALSE(result.debug_info.admissible_road_areas.empty());
 }
 
 TEST(DrivableAreaComplianceTest, ReturnsUnavailableWhenRequiredInputsAreMissing)
```
go-sakayori, line 252: Tests cover road, shoulder, hatched road marking, road-border fallback, and outside-area failure — good. Still missing:

- `intersection_areas` branch (counted in `EgoAreaEvaluation`).
- `parking_lots` branch.
- `unavailable_drivable_area_query_failed` and `unavailable_route_handler_map_not_ready` reason paths.
- The failure-case asserts non-empty `failing_corners` but doesn't pin `first_failure_time_s` or corner indices, so a regression in `append_first_failure_debug_info` ordering wouldn't be caught.

Not blocking, but easy adds now that the test harness builds a real `RouteHandler`.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/drivable_area_compliance.cpp
bskkimm: Fixed. Switched DAC cpp includes to project-source-relative `metrics/geometry/...` paths.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/drivable_area_compliance.cpp
bskkimm: Fixed. Removed the local point helper and now use `autoware_utils_geometry::create_point(...)` directly where a `Point2d` plus z is converted.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/drivable_area_compliance.hpp
bskkimm: Fixed. Switched the DAC header include to `metrics/geometry/ego_footprint.hpp`.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/drivable_area_compliance.cpp
bskkimm: Fixed. Simplified lanelet/map/line debug point conversion to use `autoware::lanelet2_utils::to_ros(...)` directly instead of converting through `to_polygon_2d`.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/drivable_area_compliance.cpp
bskkimm: Fixed. DAC now validates all footprint evaluations before filling debug info, so unavailable results do not carry partial debug payload.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/drivable_area_compliance.cpp
bskkimm: Fixed. Renamed the size-mismatch reason to `unavailable_footprint_evaluation_size_mismatch`; I also updated the PR description to state that DAC reason values changed while topic names and JSON keys remain unchanged.

planning/autoware_planning_data_analyzer/test/metrics/test_drivable_area_compliance.cpp
bskkimm: Fixed. Added tests for intersection area, parking lot, map-not-ready, missing footprint evaluation, footprint-evaluation size mismatch, and first-failure time / corner-index consistency.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/drivable_area_compliance.cpp
```diff
@@ -16,15 +16,17 @@
 
 #include "metrics/geometry/metric_utils.hpp"
 
+#include <autoware/lanelet2_utils/conversion.hpp>
 #include <autoware_utils_geometry/boost_geometry.hpp>
```
go-sakayori, line 19: Missing direct include for `autoware_utils_geometry::create_point`. The file now calls `create_point` 14 times, but `create_point` is declared in `<autoware_utils_geometry/geometry.hpp>` — only `boost_geometry.hpp` is included here. The build only works because some other header (likely `route_handler/route_handler.hpp` or `ego_footprint.hpp`) pulls it in transitively.

Google's C++ Style Guide explicitly disallows this ("Include What You Use"):

> Do not rely on transitive inclusions. This allows people to remove no-longer-needed `#include` statements from their headers without breaking clients.

Please add:

```cpp
#include <autoware_utils_geometry/geometry.hpp>
```

go-sakayori (approved):

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/drivable_area_compliance.cpp
bskkimm: Fixed. Added the direct `<autoware_utils_geometry/geometry.hpp>` include so `create_point` does not rely on transitive includes.

PR #428, opened 2026-05-19
feat(planning_data_analyzer): migrate DDC metric

bskkimm:
## Description

This is the next PR in the EPDMS patch series after the DAC migration PR.

PR type: subscore.

### Logic Change

As-is pipeline:
trajectory point metrics carried DDC-specific route/oncoming checks inline with other trajectory-point calculations.

To-be pipeline:
trajectory sample -> shared local lanelet context -> DDC score logic -> DDC reason/debug result.

The score behavior is intended to match the prepared reference DDC migration while reducing DDC-specific logic inside trajectory point metric calculation.

### Notes for reviewers

- This PR is intentionally limited to DDC migration.
- Reuses the EPDMS helper structure introduced by previous PRs.
- No debug-topic contract changes are intended in this PR.

## How was this PR tested?

- Rebased onto upstream main after PR #427 was merged.
- Changed-file pre-commit passed.
- Prepared Takanawa validation before opening: DDC reproduced the reference result with 81 non-1, 0 unavailable, and 0 changed DDC timestamps versus the prepared reference artifact.

### DIFF
--- a/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/driving_direction_compliance.cpp
+++ b/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/driving_direction_compliance.cpp
@@ -46,13 +46,21 @@ DrivingDirectionComplianceResult calculate_driving_direction_compliance(
   for (size_t i = 0; i < evaluation_points.size(); ++i) {
     const double t_i = evaluation_points.at(i).time_from_start_s;
     double horizon_sum = 0.0;
+    std::size_t first_index_in_window = i;
     for (size_t j = 0; j <= i; ++j) {
       const double t_j = evaluation_points.at(j).time_from_start_s;
       if (t_i - t_j <= params.horizon_s + 1.0e-6) {
+        first_index_in_window = j;
         horizon_sum += filtered_progress[j];
       }
     }
-    result.max_oncoming_progress_m = std::max(result.max_oncoming_progress_m, horizon_sum);
+    if (horizon_sum > result.max_oncoming_progress_m + 1.0e-9) {
+      result.max_oncoming_progress_m = horizon_sum;
+      result.worst_window_start_time_s =
+        evaluation_points.at(first_index_in_window).time_from_start_s;
+      result.worst_window_end_time_s = t_i;
+      result.worst_window_sample_count = i - first_index_in_window + 1U;
+    }
   }
 
   if (result.max_oncoming_progress_m < params.compliance_threshold_m) {
--- a/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/driving_direction_compliance.cpp
+++ b/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/driving_direction_compliance.cpp
@@ -46,21 +46,13 @@ DrivingDirectionComplianceResult calculate_driving_direction_compliance(
   for (size_t i = 0; i < evaluation_points.size(); ++i) {
     const double t_i = evaluation_points.at(i).time_from_start_s;
     double horizon_sum = 0.0;
-    std::size_t first_index_in_window = i;
     for (size_t j = 0; j <= i; ++j) {
       const double t_j = evaluation_points.at(j).time_from_start_s;
       if (t_i - t_j <= params.horizon_s + 1.0e-6) {
-        first_index_in_window = j;
         horizon_sum += filtered_progress[j];
       }
     }
-    if (horizon_sum > result.max_oncoming_progress_m + 1.0e-9) {
-      result.max_oncoming_progress_m = horizon_sum;
-      result.worst_window_start_time_s =
-        evaluation_points.at(first_index_in_window).time_from_start_s;
-      result.worst_window_end_time_s = t_i;
-      result.worst_window_sample_count = i - first_index_in_window + 1U;
-    }
+    result.max_oncoming_progress_m = std::max(result.max_oncoming_progress_m, horizon_sum);
   }
 
   if (result.max_oncoming_progress_m < params.compliance_threshold_m) {
--- a/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/driving_direction_compliance.hpp
+++ b/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/driving_direction_compliance.hpp
@@ -15,6 +15,7 @@
 #ifndef METRICS__EPDMS__SUBSCORES__DRIVING_DIRECTION_COMPLIANCE_HPP_
 #define METRICS__EPDMS__SUBSCORES__DRIVING_DIRECTION_COMPLIANCE_HPP_
 
+#include <cstddef>
 #include <limits>
 #include <string>
 #include <vector>
@@ -43,6 +44,9 @@ struct DrivingDirectionComplianceResult
   bool available{false};
   std::string reason{"unavailable"};
   double max_oncoming_progress_m{0.0};
+  double worst_window_start_time_s{0.0};
+  double worst_window_end_time_s{0.0};
+  std::size_t worst_window_sample_count{0U};
 };
 
 DrivingDirectionComplianceResult calculate_driving_direction_compliance(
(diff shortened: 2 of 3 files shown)


go-sakayori (commented):
Reviewed the DDC migration. The route-lane / intersection consolidation in `trajectory_metrics.cpp` looks like a clean follow-up to the EPDMS helper pattern, and the validation note (81 non-1, 0 changed timestamps) gives good evidence the score behavior matches the reference.

One real bug in the new worst-window bookkeeping plus a few questions about the new `worst_window_*` fields — comments inline.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/driving_direction_compliance.cpp
```diff
@@ -46,13 +46,21 @@ DrivingDirectionComplianceResult calculate_driving_direction_compliance(
   for (size_t i = 0; i < evaluation_points.size(); ++i) {
     const double t_i = evaluation_points.at(i).time_from_start_s;
     double horizon_sum = 0.0;
+    std::size_t first_index_in_window = i;
     for (size_t j = 0; j <= i; ++j) {
       const double t_j = evaluation_points.at(j).time_from_start_s;
       if (t_i - t_j <= params.horizon_s + 1.0e-6) {
+        first_index_in_window = j;
```
go-sakayori: Bug: `first_index_in_window` collapses to `i` here, so the recorded window is always size 1.

Since `j` iterates upward and `j = i` always satisfies the horizon condition, this assignment rewrites `first_index_in_window` on every match and the final value ends up equal to `i`. Consequence:

- `worst_window_start_time_s == t_i == worst_window_end_time_s`
- `worst_window_sample_count == 1`

…regardless of how many samples are actually inside the horizon window. The score (`max_oncoming_progress_m`) is unaffected, which is why the reference-result validation still passes — but the new tracking fields are wrong by construction.

Suggested fix — capture the earliest qualifying `j` only:

```cpp
std::size_t first_index_in_window = i;
bool found_first = false;
for (size_t j = 0; j <= i; ++j) {
  const double t_j = evaluation_points.at(j).time_from_start_s;
  if (t_i - t_j <= params.horizon_s + 1.0e-6) {
    if (!found_first) { first_index_in_window = j; found_first = true; }
    horizon_sum += filtered_progress[j];
  }
}
```

(Or compute the earliest in-window index ahead of the sum and skip the flag.)

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/driving_direction_compliance.hpp
```diff
@@ -43,6 +44,9 @@ struct DrivingDirectionComplianceResult
   bool available{false};
   std::string reason{"unavailable"};
   double max_oncoming_progress_m{0.0};
+  double worst_window_start_time_s{0.0};
+  double worst_window_end_time_s{0.0};
+  std::size_t worst_window_sample_count{0U};
```
go-sakayori: These three new fields are write-only in this PR — `grep` across the package shows they're populated in `driving_direction_compliance.cpp` and read nowhere (not propagated to `TrajectoryPointMetrics`, not published, not consumed by aggregation, not asserted in any test).

The PR description says "no debug-topic contract changes are intended," so I'd suggest one of:

1. Drop the fields from this PR and re-introduce them in the follow-up that actually wires them through (debug topic, aggregation, or wherever); or
2. Keep them, but fix the `first_index_in_window` bug (other comment) and add at least one test in `test_driving_direction_compliance.cpp` that asserts `worst_window_start_time_s` / `worst_window_end_time_s` / `worst_window_sample_count` over a multi-sample horizon.

As they stand they'll ship broken and silent, and the bug won't be noticed until someone tries to use them.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/driving_direction_compliance.cpp
```diff
@@ -46,13 +46,21 @@ DrivingDirectionComplianceResult calculate_driving_direction_compliance(
...
         horizon_sum += filtered_progress[j];
       }
     }
-    result.max_oncoming_progress_m = std::max(result.max_oncoming_progress_m, horizon_sum);
+    if (horizon_sum > result.max_oncoming_progress_m + 1.0e-9) {
```
go-sakayori: Minor semantic change worth confirming is intentional: the old code was `result.max_oncoming_progress_m = std::max(...)`, which means on a tie the *later* sample's window effectively wins (no-op overwrite, but if there were tracking fields they would advance). The new `>` (with `+ 1e-9`) keeps the *earliest* window that reaches the max.

For the score this is a no-op. For the new `worst_window_*` fields it determines which window gets reported when multiple windows tie. "Prefer earliest worst window on ties" seems like a reasonable choice once you start recording the window — a one-line comment to that effect would help future readers, since the `1e-9` epsilon hides the intent.

planning/autoware_planning_data_analyzer/src/metrics/trajectory_metrics.cpp
```diff
@@ -287,11 +266,12 @@ TrajectoryPointMetrics calculate_trajectory_point_metrics(
           trajectory.points.at(i - 1).pose.position, trajectory.points.at(i).pose.position);
       }
       const auto & point = trajectory.points.at(i);
+      const auto local_context = compute_driving_direction_local_context(point.pose, route_handler)
```
go-sakayori, line 269: FYI / not a blocker — this isn't a pure refactor of the old `is_pose_in_route_lane`.

The deleted helper used `route_handler->getClosestLaneletWithinRoute(...)` plus a `getRoadLaneletsAtPose` / `isRouteLanelet` fallback (closest-lanelet semantics with route filtering).

`local_context.in_route_lane_polygon` (from `compute_driving_direction_local_context` in `lanelet_queries.cpp`) is computed via actual point-in-lanelet polygon containment plus a `kAdmissibleLaneMarginM` margin over `collect_local_route_consistent_lanelets`. Those are not equivalent — a pose near a route lanelet boundary, or one whose closest route lanelet doesn't contain it, can land on different answers.

The Takanawa validation (81 non-1, 0 changed DDC timestamps vs the reference) is the right kind of evidence that this matches the intended reference behavior, so I'm fine with the change — just calling it out so future readers don't read this as a no-op refactor.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/driving_direction_compliance.cpp
bskkimm: Removed the worst-window bookkeeping from this PR, so this incorrect index tracking no longer exists. DDC now keeps only the score-relevant max oncoming progress.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/driving_direction_compliance.hpp
bskkimm: Agreed. Dropped the write-only `worst_window_*` fields from this PR. They can be reintroduced later together with the debug/topic/test wiring that consumes them.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/driving_direction_compliance.cpp
bskkimm: Restored the score-only accumulation to `std::max(...)` after removing the worst-window fields, so there is no reported-window tie behavior in this PR anymore.

planning/autoware_planning_data_analyzer/src/metrics/trajectory_metrics.cpp
bskkimm: Acknowledged. This is intentional for the DDC migration rather than a pure no-op refactor: the local context matches the prepared reference logic, and the Takanawa validation showed 0 changed DDC timestamps.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/driving_direction_compliance.hpp
```diff
@@ -15,6 +15,7 @@
 #ifndef METRICS__EPDMS__SUBSCORES__DRIVING_DIRECTION_COMPLIANCE_HPP_
 #define METRICS__EPDMS__SUBSCORES__DRIVING_DIRECTION_COMPLIANCE_HPP_
 
+#include <cstddef>
```
go-sakayori:
```suggestion
```

go-sakayori (changes requested):

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/driving_direction_compliance.hpp
bskkimm: Removed the unused `<cstddef>` include after dropping the `worst_window_*` fields.

go-sakayori (approved):

PR #429, opened 2026-05-19
feat(planning_data_analyzer): migrate TLC metric

bskkimm:
## Description

This is the next PR in the EPDMS patch series after the DDC migration PR.

PR type: subscore.

### Logic Change

As-is pipeline:
reference lanelet -> traffic-light regulatory element -> signal group -> ego footprint / stop-line overlap.

To-be pipeline:
trajectory route lanelets -> relevant traffic-light groups -> movement-filtered controlled lanelets -> signal group association -> ego polygon / stop-line overlap.

This ports the TLC logic toward the prepared stop-line based EPDMS behavior while keeping debug-topic payload out of this score-migration PR.

### Notes for reviewers

- This PR is intentionally limited to TLC migration and required turn-indicator input plumbing.
- Reuses route-lanelet and ego-footprint helpers introduced by previous EPDMS PRs.
- TLC debug payload was intentionally removed from this PR because it is not published/tested here.

## How was this PR tested?

- Rebased onto upstream main after PR #428 was merged.
- Changed-file pre-commit passed.
- Prepared Takanawa cumulative validation with metrics ['nc','dac','ddc','tlc']:
  - Run artifact: /home/beomseokkim2/rosbag/x2_takanawa/run_logs/20260520_012118
  - TLC exact match against reference: 68 non-1, 0 unavailable, 0 changed TLC timestamps.
  - DAC exact match remained 396 non-1.
  - DDC exact match remained 81 non-1.
  - NC differs only by the accepted PR #426 de-dup delta: 74 -> 87 non-1.

### DIFF
--- a/planning/autoware_planning_data_analyzer/src/bag_handler.hpp
+++ b/planning/autoware_planning_data_analyzer/src/bag_handler.hpp
@@ -72,6 +72,8 @@ struct BagData
       max_buffer_msgs);
     create_buffer<SteeringReport>(
       "/vehicle/status/steering_status", buffer_duration_sec, max_buffer_msgs);
+    create_buffer<TurnIndicatorsReport>(
+      "/vehicle/status/turn_indicators_status", buffer_duration_sec, max_buffer_msgs);
   }
 
   // Constructor with topic names
@@ -108,6 +110,10 @@ struct BagData
         topic_names.traffic_signals_topic, buffer_duration_sec, max_buffer_msgs);
     }
     create_buffer<SteeringReport>(topic_names.steering_topic, buffer_duration_sec, max_buffer_msgs);
+    if (!topic_names.turn_indicators_topic.empty()) {
+      create_buffer<TurnIndicatorsReport>(
+        topic_names.turn_indicators_topic, buffer_duration_sec, max_buffer_msgs);
+    }
   }
 
   rcutils_time_point_value_t timestamp;
@@ -232,6 +238,23 @@ struct BagData
       synchronized_data->steering_status = steer_buffer->get_closest(traj_stamp_ns, tolerance_ms);
     } else if (steer_buffer) {
       synchronized_data->steering_status = steer_buffer->get_closest(target_time, tolerance_ms);
+    }
+
+    std::shared_ptr<Buffer<TurnIndicatorsReport>> turn_indicators_buffer;
+    for (const auto & [topic, buffer] : buffers) {
+      if (auto tb = std::dynamic_pointer_cast<Buffer<TurnIndicatorsReport>>(buffer)) {
+        turn_indicators_buffer = tb;
+        break;
+      }
+    }
+    if (turn_indicators_buffer && synchronized_data->trajectory) {
+      const auto traj_stamp_ns =
+        rclcpp::Time(synchronized_data->trajectory->header.stamp).nanoseconds();
+      synchronized_data->turn_indicators_status =
+        turn_indicators_buffer->get_closest(traj_stamp_ns, tolerance_ms);
+    } else if (turn_indicators_buffer) {
+      synchronized_data->turn_indicators_status =
+        turn_indicators_buffer->get_closest(target_time, tolerance_ms);
     }
 
     // Get objects
--- a/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/traffic_light_compliance.cpp
+++ b/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/traffic_light_compliance.cpp
@@ -25,10 +25,8 @@
 #include <boost/geometry/algorithms/intersects.hpp>
 
 #include <algorithm>
-#include <cmath>
 #include <memory>
 #include <optional>
-#include <string>
 #include <unordered_map>
 #include <unordered_set>
 #include <vector>
@@ -82,41 +80,6 @@ const autoware_perception_msgs::msg::TrafficLightGroup * find_signal_group(
   return nullptr;
 }
 
-std::vector<geometry_msgs::msg::Point> polygon_to_msg_points(
-  const autoware_utils_geometry::Polygon2d & polygon, const double z)
-{
-  std::vector<geometry_msgs::msg::Point> points;
-  points.reserve(polygon.outer().size() + 1U);
-  for (const auto & point : polygon.outer()) {
-    geometry_msgs::msg::Point msg;
-    msg.x = point.x();
-    msg.y = point.y();
-    msg.z = z;
-    points.push_back(msg);
-  }
-  if (
-    !points.empty() &&
-    (points.front().x != points.back().x || points.front().y != points.back().y)) {
-    points.push_back(points.front());
-  }
-  return points;
-}
-
-std::vector<geometry_msgs::msg::Point> stop_line_to_msg_points(
-  const lanelet::ConstLineString3d & stop_line, const double z)
-{
-  std::vector<geometry_msgs::msg::Point> points;
-  points.reserve(stop_line.size());
-  for (const auto & point : stop_line) {
-    geometry_msgs::msg::Point msg;
-    msg.x = point.x();
-    msg.y = point.y();
-    msg.z = z;
-    points.push_back(msg);
-  }
-  return points;
-}
-
 std::optional<TurnDirection> infer_turn_direction_from_indicator(
   const std::shared_ptr<TurnIndicatorsReport> & turn_indicators_status)
 {
@@ -153,19 +116,6 @@ std::optional<TurnDirection> infer_turn_direction_from_route_lanelets(
     }
   }
   return inferred;
-}
-
-std::string turn_direction_to_string(const TurnDirection turn_direction)
-{
-  switch (turn_direction) {
-    case TurnDirection::Straight:
-      return "straight";
-    case TurnDirection::Left:
-      return "left";
-    case TurnDirection::Right:
-      return "right";
-  }
-  return "unknown";
 }
 
 bool matches_intended_turn_direction(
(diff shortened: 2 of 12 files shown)


planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/traffic_light_compliance.cpp
```diff
@@ -42,27 +39,174 @@ using autoware::route_handler::RouteHandler;
...
+{
+  std::vector<lanelet::Id> candidate_ids;
+  candidate_ids.reserve(1U + group.selected_lanelets.size() + group.route_lanelets.size());
+  candidate_ids.push_back(group.regulatory_element_id);
+  for (const auto & lanelet : group.selected_lanelets) {
+    candidate_ids.push_back(lanelet.id());
+  }
+  for (const auto & lanelet : group.route_lanelets) {
+    candidate_ids.push_back(lanelet.id());
+  }
```
go-sakayori: Question on the new fallback chain.

The old `find_signal_group` keyed strictly on `reg_elem->id()`. This version tries reg-elem id, then `selected_lanelets` ids, then `route_lanelets` ids. Grepping the producers in `autoware/universe` — e.g. `autoware_traffic_light_multi_camera_fusion_node.cpp:183` (`signal_out.traffic_light_group_id = reg_ele_id`) and `autoware_diffusion_planner/lane_segments.cpp:141` — shows `traffic_light_group_id` is always set to the *regulatory-element* id in production.

So this fallback only fires when (a) there's a non-standard bag where `traffic_light_group_id` was set to a lanelet id, or (b) there's a numeric collision between a `traffic_light_group_id` and an unrelated `lanelet::Id`, in which case the lookup silently false-matches and the score degrades silently.

Could you either drop the fallback (if no real data needs it) or add a short comment naming the bag variant / map encoding it handles? As written, a future reader can't justify the fallback and the silent-false-match risk is non-trivial.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/traffic_light_compliance.cpp
```diff
@@ -42,27 +39,174 @@ using autoware::route_handler::RouteHandler;
...
+      }
+      if (selected_lanelet_ids.insert(lanelet.id()).second) {
+        group.selected_lanelets.push_back(lanelet);
+      }
+    }
+
+    if (group.selected_lanelets.empty()) {
+      group.selected_lanelets = group.route_lanelets;
+    }
+  }
```
go-sakayori, line 184: Design question on turn-indicator scope.

`turn_indicators_status` is sampled once at the trajectory header timestamp (via `bag_handler.hpp:243-258`). If it reports `ENABLE_LEFT`, this loop assigns `intended_turn_direction = Left` to *every* `RelevantTrafficLightGroup` in the trajectory — not just the upcoming one. For a trajectory that spans multiple intersections, that would over-filter `selected_lanelets` at the later intersections to "left-turn only," which is wrong if the driver only intends to turn left at the first.

Probably tolerable for current Takanawa scenes (single intersection per trajectory horizon), but conceptually fragile. Two reasonable options:

1. Apply the indicator only to the nearest group along the route and fall through to `infer_turn_direction_from_route_lanelets` for later groups; or
2. Document this limitation in a comment so it's not load-bearing for future use.

Not blocking — just want to make sure this matches the prepared reference's intent.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/traffic_light_compliance.hpp
```diff
@@ -39,7 +44,10 @@ TrafficLightComplianceResult calculate_traffic_light_compliance(
   const autoware_planning_msgs::msg::Trajectory & trajectory,
   const std::shared_ptr<TrafficLightGroupArray> & traffic_signals,
   const std::shared_ptr<RouteHandler> & route_handler,
-  const autoware::vehicle_info_utils::VehicleInfo & vehicle_info);
+  const autoware::vehicle_info_utils::VehicleInfo & vehicle_info,
+  const std::shared_ptr<TurnIndicatorsReport> & turn_indicators_status = nullptr,
+  const std::vector<TrajectoryFootprintEvaluation> * evaluations = nullptr,
+  const lanelet::ConstLanelets * route_relevant_lanelets = nullptr);
```
go-sakayori, line 50: Dead optional plumbing.

The only caller (`trajectory_metrics.cpp:298-300`) passes `&footprint_evaluations` but never passes `route_relevant_lanelets`, so the function always falls back to `collect_route_relevant_lanelets(trajectory, route_handler)` at `.cpp:244`. Same shape of issue as the `worst_window_*` fields in PR #428 v1 — preemptive plumbing with no live caller.

Two options:

1. Drop this parameter from the signature until something actually exercises it; or
2. Wire the cache through at the callsite. DAC (`calculate_drivable_area_compliance`) computes route-relevant lanelets too — if you compute them once in `trajectory_metrics.cpp` and pass them into both DAC and TLC, the parameter starts paying for itself.

If (2) is the plan but it's just out of scope for this PR, fine — but at minimum consider dropping the unused parameter for now.

planning/autoware_planning_data_analyzer/test/metrics/test_traffic_light_compliance.cpp
```diff
@@ -25,3 +25,25 @@ TEST(TrafficLightComplianceTest, EmptyTrajectoryIsUnavailable)
...
+  vehicle_info.rear_overhang_m = 1.0;
+  vehicle_info.left_overhang_m = 0.9;
+  vehicle_info.right_overhang_m = 0.9;
+
+  const auto result = autoware::planning_data_analyzer::metrics::calculate_traffic_light_compliance(
+    trajectory, nullptr, nullptr, vehicle_info);
+
+  EXPECT_FALSE(result.available);
+  EXPECT_EQ(result.reason, "unavailable_no_route_handler");
+}
```
go-sakayori, line 156: Test coverage gap (not a blocker, but worth a look before merge).

The new test only exercises an early-return error path. None of the new code paths are unit-tested:

- `build_relevant_traffic_light_groups` — multi-lanelet grouping per reg-elem, stop-line capture.
- `infer_turn_direction_from_indicator` — `ENABLE_LEFT` / `ENABLE_RIGHT` / `NONE`.
- `infer_turn_direction_from_route_lanelets` — mixed-direction → `std::nullopt`.
- `matches_intended_turn_direction` — lanelet without `turn_direction` tag → `false`.
- `find_signal_group` fallback chain (if it's kept after the other comment).
- `evaluations` reuse path matches non-reuse output for the same input.

Takanawa reference (68 non-1) is the right kind of integration evidence, but unit tests would catch regressions in each of these in isolation. At minimum, a happy-path test that exercises `build_relevant_traffic_light_groups` (the most non-trivial new logic) would be valuable.

planning/autoware_planning_data_analyzer/src/bag_handler.hpp
```diff
@@ -72,6 +72,8 @@ struct BagData
       max_buffer_msgs);
     create_buffer<SteeringReport>(
       "/vehicle/status/steering_status", buffer_duration_sec, max_buffer_msgs);
+    create_buffer<TurnIndicatorsReport>(
+      "/vehicle/status/turn_indicators_status", buffer_duration_sec, max_buffer_msgs);
```
go-sakayori, line 76: Minor / pre-existing pattern — non-blocking.

This default-ctor branch creates the turn-indicators buffer unconditionally, while the topic-names ctor at line 113 guards with `if (!topic_names.turn_indicators_topic.empty())`. The pattern matches existing behavior for traffic_signals and steering, so I don't think it needs changing here, but the inconsistency surfaces in tests (`test_bag_handler.cpp` had to bump 9 → 10).

Worth a one-line internal note (in the PR description or a follow-up tracking ticket) that the default ctor's buffer set is intentionally always-on and isn't config-gated.

go-sakayori (commented):
Reviewed the TLC migration. The route-lanelet + turn-direction approach looks like a real improvement over the per-point reference-lanelet lookup, and the Takanawa reference match (68 non-1, 0 changed timestamps) is good evidence the score behavior is intact.

A few questions / concerns inline, in roughly descending order of importance. Also a meta-comment: this PR isn't strictly a `subscore` change — it adds `TurnIndicatorsReport` topic plumbing (config, bag-handler buffer, sync-data field). Worth calling out in the description that this is bundled.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/traffic_light_compliance.cpp
```diff
@@ -42,27 +39,174 @@ using autoware::route_handler::RouteHandler;
...
+  const auto indicator_turn_direction = infer_turn_direction_from_indicator(turn_indicators_status);
+  std::unordered_map<lanelet::Id, std::size_t> index_by_reg_elem_id;
+  for (const auto & lanelet : route_lanelets) {
+    if (!route_handler->isRoadLanelet(lanelet)) {
+      continue;
+    }
+    for (const auto & reg_elem :
+         lanelet.regulatoryElementsAs<lanelet::autoware::AutowareTrafficLight>()) {
+      const auto [it, inserted] =
+        index_by_reg_elem_id.emplace(reg_elem->id(), index_by_reg_elem_id.size());
```
go-sakayori, line 138: Style nit (combined) — non-blocking.

Two small things in this block:

1. `emplace(reg_elem->id(), index_by_reg_elem_id.size())` is a subtle idiom: the size-before-insert captures the index that `groups.push_back(group)` below will assign to the new element. It's correct, but easy to miss on a quick read. A one-line `// Use the map size before insertion as the index assigned to the new group.` above this line would help future readers.
2. `groups.at(it->second)` at line 166: `at()` does range-checked access; here `it->second` is the index we just (or earlier) reserved for this group, so it's always valid. `groups[it->second]` reads cleaner. Pure nit.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/traffic_light_compliance.cpp
bskkimm: Dropped the fallback chain. `find_signal_group` now matches only `traffic_light_group_id == regulatory_element_id`, so it no longer silently matches lanelet IDs.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/traffic_light_compliance.cpp
bskkimm: Added a code comment documenting the current scope: the indicator is sampled once at the trajectory timestamp and applied to all groups. Per-group intent inference is left as a follow-up if multi-intersection horizons become relevant.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/traffic_light_compliance.hpp
bskkimm: Wired the route-lanelet cache through from `calculate_trajectory_point_metrics()` to TLC, so this optional input now has a live caller. I kept DAC signature unchanged in this PR to avoid broadening the review scope.

planning/autoware_planning_data_analyzer/test/metrics/test_traffic_light_compliance.cpp
bskkimm: Added a happy-path unit test that builds a minimal route lanelet with an `AutowareTrafficLight` regulatory element and verifies crossing its red stop line returns `red_light_stop_line_crossed`.

planning/autoware_planning_data_analyzer/src/bag_handler.hpp
bskkimm: Left this unchanged because it matches the existing default-constructor pattern for always-on buffers. I will keep the config-gated cleanup as follow-up scope rather than mix it into this TLC migration PR.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/traffic_light_compliance.cpp
bskkimm: Applied both nits: added a short comment explaining the size-before-insert index idiom and changed `groups.at(it->second)` to `groups[it->second]`.

go-sakayori (approved):

PR #431, opened 2026-05-20
feat(planning_data_analyzer): migrate TTC metric

bskkimm:
## Description
Fourth PR in the EPDMS metric migration series.

Migrates TTC within bound to the NAVSIM-style logged-object future check while reusing the shared EPDMS helpers introduced in earlier PRs.

Pipeline change:
- As-is: TTC used predicted object paths from the synchronized object message and treated the forward ego-frame half-plane as ahead.
- To-be: TTC uses logged future tracked-object states, projects ego at 0.0/0.3/0.6/0.9 s, ignores unknown and previously-collided objects, applies the NAVSIM/nuPlan ahead cone instead of the full forward half-plane, and uses the NAVSIM ahead / bad-area-or-intersection / behind logic.

## Validation
- pre-commit run on changed files: passed
- Takanawa full run through NC/DAC/DDC/TLC/TTC using pilot-auto.x2 underlay: completed
- TTC reproduced baseline exactly: 102 non-1, 0 unavailable, same mean
- DAC/DDC/TLC unchanged vs baseline

## Notes
- Debug topic changes are intentionally reserved for a later debug-topic PR.
- Local validation used the known pilot-auto.x2 lanelet compatibility patch only during validation; it was restored before this PR branch.

### DIFF
--- a/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ttc_within_bound.cpp
+++ b/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ttc_within_bound.cpp
@@ -17,10 +17,8 @@
 #include "metrics/geometry/ego_footprint.hpp"
 #include "metrics/geometry/lanelet_queries.hpp"
 #include "metrics/geometry/metric_utils.hpp"
+#include "metrics/geometry/object_tracks.hpp"
 
-#include <autoware_utils_geometry/boost_geometry.hpp>
-#include <autoware_utils_geometry/boost_polygon_utils.hpp>
-#include <autoware_utils_geometry/geometry.hpp>
 #include <tf2/LinearMath/Vector3.hpp>
 
 #include <boost/geometry.hpp>
@@ -30,8 +28,8 @@
 #include <cmath>
 #include <limits>
 #include <memory>
-#include <optional>
 #include <string>
+#include <unordered_set>
 #include <vector>
 
 namespace autoware::planning_data_analyzer::metrics
@@ -42,20 +40,10 @@ using autoware::route_handler::RouteHandler;
 namespace
 {
 
-using autoware_utils_geometry::LinearRing2d;
-using autoware_utils_geometry::Point2d;
-using autoware_utils_geometry::Polygon2d;
-
 constexpr double kStoppedSpeedThreshold = 5.0e-3;
 constexpr std::array<double, 4> kFutureProjectionOffsetsSec{{0.0, 0.3, 0.6, 0.9}};
-
-struct PredictedObjectCache
-{
-  const autoware_perception_msgs::msg::PredictedObject * object{};
-  const autoware_perception_msgs::msg::PredictedPath * path{};
-  double dt{0.0};
-  double max_time{0.0};
-};
+constexpr double kAheadAngleThresholdRad = M_PI / 6.0;
+constexpr double kBehindAngleThresholdRad = 5.0 * M_PI / 6.0;
 
 tf2::Vector3 get_velocity_in_world_coordinate(
   const autoware_planning_msgs::msg::TrajectoryPoint & point)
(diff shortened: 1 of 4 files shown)


go-sakayori (commented):
Reviewed the TTC migration. The switch from `PredictedObjects` + predicted-paths to logged-future-tracked-objects, with UUID-based dedup of already-collided agents and reuse of the shared EPDMS footprint/area helpers, looks coherent. Takanawa baseline match (102 non-1, same mean) gives good integration evidence for the prepared NAVSIM reference.

A few questions / asks inline — mostly around documenting deliberate semantic changes and one missing test for a protective branch.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ttc_within_bound.cpp
```diff
@@ -157,43 +122,87 @@ TTCWithinBoundResult calculate_ttc_within_bound(
...
+      multiple_lanes = flags.multiple_lanes;
+      non_drivable_area = flags.non_drivable_area;
+    }
+    const bool ego_in_intersection =
+      !evaluations.empty() && index < evaluations.size() &&
+          evaluations.at(index).ego_area_evaluation.has_value() &&
+          evaluations.at(index).ego_area_evaluation->flags.intersection_context_available
+        ? evaluations.at(index).ego_area_evaluation->flags.in_intersection
+        : is_pose_in_intersection(point.pose, route_handler);
+    const bool bad_or_intersection = multiple_lanes || non_drivable_area || ego_in_intersection;
```
go-sakayori, line 172: The "lateral overlap counts as infraction unless behind" rule used to be gated only on `ego_in_intersection`. This PR widens it to `multiple_lanes || non_drivable_area || ego_in_intersection`.

The two new conditions come from `EgoAreaFlags` populated by the shared `footprint_evaluations` pass, so reading TTC alone it's not obvious what counts as "bad area." Could you add a brief comment here naming the three conditions (something like `// "bad area": straddling lanes, off drivable area, or inside an intersection`) so a future reader doesn't have to chase the flag's origin? The PR description mentions "NAVSIM ahead / bad-area-or-intersection / behind logic" — making that explicit in code as well would help.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ttc_within_bound.cpp
```diff
@@ -77,75 +65,52 @@ geometry_msgs::msg::Pose project_pose(
...
-    std::min(static_cast<std::size_t>(query_time_s / cache.dt), cache.path->path.size() - 2);
-  const double t_i = static_cast<double>(index) * cache.dt;
-  const double ratio = std::clamp((query_time_s - t_i) / cache.dt, 0.0, 1.0);
-  return autoware_utils_geometry::calc_interpolated_pose(
-    cache.path->path.at(index), cache.path->path.at(index + 1), ratio);
+bool is_agent_behind_nuplan(
+  const geometry_msgs::msg::Pose & ego_pose, const geometry_msgs::msg::Pose & object_pose)
+{
+  return relative_agent_angle(ego_pose, object_pose) > kBehindAngleThresholdRad;
 }
```
go-sakayori, line 94: Worth surfacing in the PR description: this is a deliberate behavioral narrowing vs the prior upstream main, not a pure refactor.

The deleted `is_agent_ahead` returned `forward_offset_in_ego_frame(...) > 0` — i.e. anything in the forward half-plane. `is_agent_ahead_nuplan` requires angle < π/6 (a ±30° forward cone). An object 5 m ahead and 3 m laterally has angle ≈ atan2(3, 5) ≈ 31° → previously "ahead," now NOT ahead. It would only trigger an infraction if the new `bad_or_intersection` path fires.

This is the NAVSIM parity the PR is aiming for, so I'm not asking to change the logic — just please call this out in the PR description so anyone comparing future TTC numbers against the pre-PR upstream behavior knows the cone change is intentional. Validation against the prepared reference doesn't catch this (the reference already has the cone).

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ttc_within_bound.cpp
```diff
@@ -157,43 +122,87 @@ TTCWithinBoundResult calculate_ttc_within_bound(
...
+    const bool bad_or_intersection = multiple_lanes || non_drivable_area || ego_in_intersection;
+    const double time_s = rclcpp::Duration(point.time_from_start).seconds();
 
     for (const double future_offset_s : kFutureProjectionOffsetsSec) {
-      const double query_time_s =
-        rclcpp::Duration(point.time_from_start).seconds() + future_offset_s;
+      const double query_time_s = time_s + future_offset_s;
+      const auto query_time = trajectory_start_time + rclcpp::Duration::from_seconds(query_time_s);
       const auto projected_pose = project_pose(point.pose, velocity_world, future_offset_s);
       const auto ego_polygon = create_pose_footprint(projected_pose, local_footprint);
```
go-sakayori, line 181: TTC necessarily builds its own ego polygon here because `projected_pose` differs from `point.pose` for every non-zero `future_offset_s`. The `footprint_evaluations` parameter the function accepts is therefore used only for the `EgoAreaFlags` (multiple_lanes / non_drivable_area / in_intersection / intersection_context_available) at the trajectory point — not for the polygon used in the intersection check below.

A one-line comment to that effect would prevent a future reader from "optimizing" by reusing `evaluations.at(index).ego_polygon`, which would silently use the unprojected polygon and break the future-projection geometry. E.g.:

```cpp
// footprint_evaluations is consumed above for the per-point area flags only; the polygon
// used for the intersection check below is rebuilt from projected_pose at each future offset.
```

planning/autoware_planning_data_analyzer/test/metrics/test_ttc_within_bound.cpp
```diff
@@ -111,60 +166,110 @@ TEST(TTCWithinBound, EmptyObjectsPasses)
...
+  auto objects_t1 = std::make_shared<TrackedObjects>();
+  objects_t1->header.stamp = make_stamp(1.0);
+  objects_t1->objects.push_back(make_box_object(6.0, 0.0, object_id));
+
+  const std::vector<TimedTrackedObjects> future_objects = {
+    TimedTrackedObjects{rclcpp::Time(objects_t0->header.stamp), objects_t0},
+    TimedTrackedObjects{rclcpp::Time(objects_t1->header.stamp), objects_t1}};
 
-  const auto result = calculate_ttc_within_bound(trajectory, objects, make_vehicle_info());
+  const auto result = calculate_ttc_within_bound(trajectory, future_objects, make_vehicle_info());
```
go-sakayori, line 287: Missing test for the protective `!behind` guard inside the bad-area branch.

The new test suite covers:
- `BehindCollisionDoesNotFail` — behind in *normal* area (doesn't enter the bad-area branch, so the `!behind` clause is never reached).
- `BadAreaAllowsLateralOverlapToFail` — lateral in bad area fires.

What's missing is the property the `!behind` guard at `ttc_within_bound.cpp:205` provides: "object behind ego AND bad area → still NOT an infraction." Without this, a future change that drops `!behind` would pass CI silently and only show up in integration.

Suggested case: an object at `(-4.0, 0.0)` (the existing `BehindCollisionDoesNotFail` geometry) plus `make_footprint_evaluations(..., false, true)` — should still return `available` / score=1.0.

Unrelated nit while I'm here: PR #429's TLC test uses `autoware::vehicle_info_utils::createVehicleInfo(...)` factory instead of populating fields by hand. Aligning `make_vehicle_info()` at the top of this file with that pattern would make the test setup a one-liner. Non-blocking.

planning/autoware_planning_data_analyzer/src/metrics/trajectory_metrics.cpp
```diff
@@ -232,19 +232,27 @@ TrajectoryPointMetrics calculate_trajectory_point_metrics(
...
 
-  const auto ttc_within_bound =
-    calculate_ttc_within_bound(trajectory, sync_data->objects, vehicle_info, route_handler);
+  const auto route_relevant_lanelets =
+    route_handler && route_handler->isHandlerReady()
+      ? collect_route_relevant_lanelets(trajectory, route_handler)
+      : lanelet::ConstLanelets{};
+  const auto footprint_evaluations = evaluate_trajectory_footprints(
+    trajectory, vehicle_info, route_handler,
+    route_relevant_lanelets.empty() ? nullptr : &route_relevant_lanelets, true);
```
go-sakayori, line 241: FYI / not blocking — this moves `evaluate_trajectory_footprints(..., evaluate_intersection_context=true)` to run unconditionally, where previously footprint evaluations were computed without intersection context and only inside the `route_handler-ready` else branch lower down.

The consolidation is the right call (TTC/DAC/TLC now share one pass), but two things to confirm:

1. Cost: `evaluate_intersection_context=true` adds per-point map queries that DAC and TLC don't need. Did the Takanawa run capture wall-clock to confirm overhead is acceptable? If not, an order-of-magnitude check would be reassuring.
2. The no-route case now still runs `evaluate_trajectory_footprints` with `route_handler=null` and `route_relevant_lanelets=nullptr`. That should still produce empty/default `EgoAreaFlags`, but worth confirming it doesn't trip an assertion path inside the helper for null inputs.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ttc_within_bound.cpp
bskkimm: Added a local code comment spelling out the NAVSIM bad-area branch as straddling lanes, outside drivable area, or inside intersection.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ttc_within_bound.cpp
bskkimm: Updated the PR description to explicitly call out the behavior change from the forward ego-frame half-plane to the NAVSIM/nuPlan ahead cone.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ttc_within_bound.cpp
bskkimm: Added a code comment clarifying that shared footprint evaluations are used only for current-point area flags and that TTC rebuilds the ego polygon at each projected future pose.

planning/autoware_planning_data_analyzer/test/metrics/test_ttc_within_bound.cpp
bskkimm: Added `BadAreaBehindCollisionDoesNotFail` to cover the protective `!behind` guard in the bad-area branch. I left the vehicle-info factory cleanup out of this patch to keep the review response focused and avoid unrelated churn.

planning/autoware_planning_data_analyzer/src/metrics/trajectory_metrics.cpp
bskkimm: Confirmed by validation: the Takanawa cumulative run completed with acceptable runtime and reproduced TTC exactly against the prepared baseline. The null-route path is also guarded by the footprint helper and produces default area flags rather than asserting. No code change needed here.

go-sakayori (approved):

PR #432, opened 2026-05-20
feat(planning_data_analyzer): migrate LK metric

bskkimm:
## Description

This is the next PR in the EPDMS patch series after TTC (#431). It migrates the lane keeping (LK) subscore to the prepared EPDMS implementation.

Pipeline:

- As-is: LK failed on a continuous centerline deviation run using the current reference lanelet/intersection handling, without explicit driver-signal or queue exemptions.
- To-be: LK samples the trajectory, computes lateral distance to the route reference lanelet centerline, skips turn/hazard lane-change windows, queue/queue-release periods, and intersection samples, then fails only when non-exempt deviation continues long enough.

This PR also adds the signal history buffering needed by LK and reuses the shared local intersection context instead of adding separate lanelet-query logic.

Duplication audit: performed; signal-window collection/merging and signal-history synchronization are centralized; remaining repeated turn/hazard calls are intentional because they bind different message types and predicates.

## Validation

- `colcon build --packages-select autoware_planning_data_analyzer --cmake-args -DCMAKE_BUILD_TYPE=Release`
- `colcon test --packages-select autoware_planning_data_analyzer --ctest-args -R test_metrics --output-on-failure --return-code-on-test-failure`
- `pre-commit run --all-files`
- Takanawa cumulative full run through LK, compared with baseline `/home/beomseokkim2/rosbag/x2_takanawa/run_logs/20260506_155345`

Full-run comparison:

- Samples: 8695 -> 8695
- NC: 74 -> 87 non-1 (accepted NC migration delta)
- DAC: 396 -> 396 non-1
- DDC: 81 -> 81 non-1
- TLC: 68 -> 68 non-1
- TTC: 102 -> 102 non-1
- LK: 367 -> 367 non-1

Generated full-run artifact `/home/beomseokkim2/rosbag/x2_takanawa/run_logs/20260520_203441` was deleted after comparison.

### DIFF
--- a/planning/autoware_planning_data_analyzer/config/planning_data_analyzer.param.yaml
+++ b/planning/autoware_planning_data_analyzer/config/planning_data_analyzer.param.yaml
@@ -30,6 +30,7 @@
     tf_topic: "/tf"
     acceleration_topic: "/localization/acceleration"
     steering_topic: "/vehicle/status/steering_status"
+    hazard_lights_topic: "/vehicle/status/hazard_lights_status"
     turn_indicators_topic: "/vehicle/status/turn_indicators_status"
     control_mode_topic: "/vehicle/status/control_mode"
 
--- a/planning/autoware_planning_data_analyzer/config/planning_data_analyzer.param.yaml
+++ b/planning/autoware_planning_data_analyzer/config/planning_data_analyzer.param.yaml
@@ -74,6 +74,8 @@
       lane_keep:
         max_lateral_deviation: 0.5  # Maximum absolute lateral deviation allowed for LK [m]
         max_continuous_violation_time: 2.0  # Continuous over-threshold duration allowed for LK [s]
+        lane_change_pre_grace_time: 1.0  # Time before turn/hazard activation exempted from LK [s]
+        lane_change_post_grace_time: 1.0  # Time after turn/hazard deactivation exempted from LK [s]
       override:
         # Time window [s] after every AUTONOMOUS->MANUAL transition used to compute
         # the override_only/* aggregate statistics in the summary JSON.
--- a/planning/autoware_planning_data_analyzer/src/bag_handler.hpp
+++ b/planning/autoware_planning_data_analyzer/src/bag_handler.hpp
@@ -36,6 +36,7 @@ struct BagData
   std::string trajectory_topic_key{};
   std::string candidate_trajectories_topic_key{};
   std::string gt_trajectory_topic_key{};
+  double trajectory_evaluation_horizon_s{0.0};
   // Template helper to create and configure buffer
   template <typename MessageType>
   void create_buffer(
@@ -72,6 +73,8 @@ struct BagData
       max_buffer_msgs);
     create_buffer<SteeringReport>(
       "/vehicle/status/steering_status", buffer_duration_sec, max_buffer_msgs);
+    create_buffer<HazardLightsReport>(
+      "/vehicle/status/hazard_lights_status", buffer_duration_sec, max_buffer_msgs);
     create_buffer<TurnIndicatorsReport>(
       "/vehicle/status/turn_indicators_status", buffer_duration_sec, max_buffer_msgs);
   }
@@ -80,7 +83,8 @@ struct BagData
   explicit BagData(
     const rcutils_time_point_value_t timestamp, const TopicNames & topic_names,
     const double buffer_duration_sec = 20.0, const size_t max_buffer_msgs = 10000)
-  : timestamp{timestamp}
+  : trajectory_evaluation_horizon_s{topic_names.trajectory_evaluation_horizon_s},
+    timestamp{timestamp}
   {
     // Create buffers using provided topic names
     create_buffer<TFMessage>(topic_names.tf_topic, buffer_duration_sec, max_buffer_msgs);
@@ -110,6 +114,10 @@ struct BagData
         topic_names.traffic_signals_topic, buffer_duration_sec, max_buffer_msgs);
     }
     create_buffer<SteeringReport>(topic_names.steering_topic, buffer_duration_sec, max_buffer_msgs);
+    if (!topic_names.hazard_lights_topic.empty()) {
+      create_buffer<HazardLightsReport>(
+        topic_names.hazard_lights_topic, buffer_duration_sec, max_buffer_msgs);
+    }
     if (!topic_names.turn_indicators_topic.empty()) {
       create_buffer<TurnIndicatorsReport>(
         topic_names.turn_indicators_topic, buffer_duration_sec, max_buffer_msgs);
@@ -240,6 +248,34 @@ struct BagData
       synchronized_data->steering_status = steer_buffer->get_closest(target_time, tolerance_ms);
     }
 
+    const auto synchronize_signal_status =
+      [&](const auto & signal_buffer, auto & status, auto & history) {
+        if (signal_buffer && synchronized_data->trajectory) {
+          const auto traj_stamp_ns =
+            rclcpp::Time(synchronized_data->trajectory->header.stamp).nanoseconds();
+          status = signal_buffer->get_closest(traj_stamp_ns, tolerance_ms);
+          const auto history_start_ns =
+            traj_stamp_ns - static_cast<rcutils_time_point_value_t>(2.0e9);
+          const auto history_end_ns =
+            traj_stamp_ns + static_cast<rcutils_time_point_value_t>(
+                              (trajectory_evaluation_horizon_s + 2.0) * 1.0e9);
+          history = signal_buffer->get_range(history_start_ns, history_end_ns);
+        } else if (signal_buffer) {
+          status = signal_buffer->get_closest(target_time, tolerance_ms);
+        }
+      };
+
+    std::shared_ptr<Buffer<HazardLightsReport>> hazard_lights_buffer;
+    for (const auto & [topic, buffer] : buffers) {
+      if (auto hb = std::dynamic_pointer_cast<Buffer<HazardLightsReport>>(buffer)) {
+        hazard_lights_buffer = hb;
+        break;
+      }
+    }
+    synchronize_signal_status(
+      hazard_lights_buffer, synchronized_data->hazard_lights_status,
+      synchronized_data->hazard_lights_history);
+
     std::shared_ptr<Buffer<TurnIndicatorsReport>> turn_indicators_buffer;
     for (const auto & [topic, buffer] : buffers) {
       if (auto tb = std::dynamic_pointer_cast<Buffer<TurnIndicatorsReport>>(buffer)) {
@@ -247,15 +283,9 @@ struct BagData
         break;
       }
     }
-    if (turn_indicators_buffer && synchronized_data->trajectory) {
-      const auto traj_stamp_ns =
-        rclcpp::Time(synchronized_data->trajectory->header.stamp).nanoseconds();
-      synchronized_data->turn_indicators_status =
-        turn_indicators_buffer->get_closest(traj_stamp_ns, tolerance_ms);
-    } else if (turn_indicators_buffer) {
-      synchronized_data->turn_indicators_status =
-        turn_indicators_buffer->get_closest(target_time, tolerance_ms);
-    }
+    synchronize_signal_status(
+      turn_indicators_buffer, synchronized_data->turn_indicators_status,
+      synchronized_data->turn_indicators_history);
 
     // Get objects
     // Find objects buffer
--- a/planning/autoware_planning_data_analyzer/src/bag_handler.hpp
+++ b/planning/autoware_planning_data_analyzer/src/bag_handler.hpp
@@ -31,6 +31,8 @@
 namespace autoware::planning_data_analyzer
 {
 
+constexpr rcutils_time_point_value_t kSignalHistoryPaddingNs = 2'000'000'000;
+
 struct BagData
 {
   std::string trajectory_topic_key{};
@@ -254,11 +256,11 @@ struct BagData
           const auto traj_stamp_ns =
             rclcpp::Time(synchronized_data->trajectory->header.stamp).nanoseconds();
           status = signal_buffer->get_closest(traj_stamp_ns, tolerance_ms);
-          const auto history_start_ns =
-            traj_stamp_ns - static_cast<rcutils_time_point_value_t>(2.0e9);
+          const auto history_start_ns = traj_stamp_ns - kSignalHistoryPaddingNs;
           const auto history_end_ns =
-            traj_stamp_ns + static_cast<rcutils_time_point_value_t>(
-                              (trajectory_evaluation_horizon_s + 2.0) * 1.0e9);
+            traj_stamp_ns +
+            static_cast<rcutils_time_point_value_t>(trajectory_evaluation_horizon_s * 1.0e9) +
+            kSignalHistoryPaddingNs;
           history = signal_buffer->get_range(history_start_ns, history_end_ns);
         } else if (signal_buffer) {
           status = signal_buffer->get_closest(target_time, tolerance_ms);
--- a/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/lane_keeping.cpp
+++ b/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/lane_keeping.cpp
@@ -14,6 +14,7 @@
 
 #include "lane_keeping.hpp"
 
+#include <algorithm>
 #include <cmath>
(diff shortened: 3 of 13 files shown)


go-sakayori (commented):
Reviewed the LK migration. The hazard-lights plumbing follows the same shape as PR #429's turn-indicators wiring, the new `Buffer<T>::get_range` is a reasonable helper, and the cumulative Takanawa numbers (LK 367 → 367, others unchanged) give good evidence the migration is score-stable.

Three real findings inline, in roughly descending order of importance.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/lane_keeping.cpp
```diff
@@ -14,48 +14,107 @@
...
+    double progress_window_distance = 0.0;
+    for (std::size_t lookback = index; lookback > 0; --lookback) {
+      const auto & previous = evaluation_points.at(lookback - 1U);
+      const auto & current = evaluation_points.at(lookback);
+      if (current.time_from_start.seconds() < progress_window_start) {
+        break;
+      }
+      progress_window_distance =
+        evaluation_point.cumulative_progress_m - previous.cumulative_progress_m;
+    }
```
go-sakayori: I think this loop has an off-by-one that makes `progress_window_distance` overestimate.

The assignment uses `previous = evaluation_points.at(lookback - 1U)`. When the loop breaks (because `current.time_from_start.seconds() < progress_window_start`), the *last* successful assignment was made when `lookback` was one step higher — i.e. `current` was still in the window. That means `previous` was at one index *before* the earliest in-window point — a point whose timestamp is BELOW `progress_window_start`. So the distance reported is `ep[index].cum - ep[earliest_in_window - 1].cum`, not `ep[index].cum - ep[earliest_in_window].cum`.

Concrete impact: `queue_exempt` requires `progress_window_distance <= queue_progress_threshold`. With the overestimate, points that are genuinely queueing get falsely classified as moving, so the queue exemption fires less often than intended. The existing unit tests don't catch this because they use sparse evaluation points where the off-by-one doesn't flip the boolean. The 367 → 367 baseline match means either the prepared reference has the same off-by-one, or the per-sample increment is too small to change the score at production sample rates.

Fix is one line — measure from `current` instead:

```cpp
progress_window_distance =
  evaluation_point.cumulative_progress_m - current.cumulative_progress_m;
```

…which then needs `current` to be the smallest in-window index when the loop exits. Easier to read as a single backwards scan that records the earliest in-window index up front, then subtracts once:

```cpp
std::size_t earliest_in_window = index;
while (earliest_in_window > 0 &&
       evaluation_points.at(earliest_in_window - 1U).time_from_start.seconds() >= progress_window_start) {
  --earliest_in_window;
}
const double progress_window_distance =
  evaluation_point.cumulative_progress_m -
  evaluation_points.at(earliest_in_window).cumulative_progress_m;
```

As a side benefit that's also O(n) total across the outer loop rather than O(n²) (since `earliest_in_window` only moves forward across outer iterations — a two-pointer sliding window). Not critical for n≈40, but cleaner.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/lane_keeping.hpp
```diff
@@ -26,13 +28,29 @@ struct LaneKeepingParameters
 {
   double max_lateral_deviation{0.5};
   double max_continuous_violation_time{2.0};
+  double lane_change_pre_grace_time{0.5};
+  double lane_change_post_grace_time{1.0};
```
go-sakayori: These two parameters are declared with defaults but never read anywhere — `grep` across the package finds only this declaration.

The call site at `trajectory_metrics.cpp:448, 451` hardcodes `1.0, 1.0` instead of forwarding `parameters.lane_change_pre_grace_time` / `parameters.lane_change_post_grace_time`. Result: changing `lane_change_pre_grace_time` in the YAML config silently does nothing, which is a real config footgun.

Same shape of issue as PR #428 v1's `worst_window_*` fields and PR #429's `route_relevant_lanelets` parameter. Two ways out:

1. Wire the params through (likely the intent given the name match) by passing `lane_keeping_params.lane_change_pre_grace_time` / `..._post_grace_time` into the two `collect_signal_active_windows` calls.
2. Drop the fields until they're consumed.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/lane_keeping.hpp
```diff
@@ -26,13 +28,29 @@ struct LaneKeepingParameters
...
+  double cumulative_progress_m{std::numeric_limits<double>::quiet_NaN()};
+};
+
+struct LaneKeepingResult
+{
+  double score{0.0};
+  double first_failure_time_s{std::numeric_limits<double>::infinity()};
+  double max_continuous_violation_time_s{0.0};
+  double peak_abs_lateral_deviation_m{0.0};
 };
```
go-sakayori: These three new fields are populated by `calculate_lane_keeping_result` but no caller reads them — `grep` shows the only invocation site is the `calculate_lane_keeping_score` wrapper, which extracts `.score`. The new tests (`LaneChangeWindowBreaksViolationRun`, `QueueExemptionAndReleaseGraceBreakViolationRun`) also assert only the score.

This is the same pattern as PR #428 v1's `worst_window_*` fields, which was eventually resolved by dropping them in the fix-up commit. The DAC subscore has analogous fields (`debug_info.first_failure_time_s`) but those have test coverage at `test_drivable_area_compliance.cpp:241-242`, so the precedent here is "if the field exists, a test asserts it."

Three options:

1. Wire the fields through to `TrajectoryPointMetrics` / debug output (probably the eventual plan given the field names mirror DAC's debug info).
2. Drop them until a consumer lands.
3. At minimum, add tests that exercise the field values — e.g. `EXPECT_GT(result.peak_abs_lateral_deviation_m, ...)` in `ReturnsZeroForContinuousViolationLongerThanWindow`, plus `EXPECT_DOUBLE_EQ(result.first_failure_time_s, ...)` on the failing scenario.

planning/autoware_planning_data_analyzer/src/metrics/trajectory_metrics.cpp
```diff
@@ -350,39 +419,55 @@ TrajectoryPointMetrics calculate_trajectory_point_metrics(
...
+    auto turn_windows = collect_signal_active_windows<TurnIndicatorsReport>(
+      sync_data->turn_indicators_history, trajectory_start_time,
+      [](const auto & msg) {
+        return msg.report == TurnIndicatorsReport::ENABLE_LEFT ||
+               msg.report == TurnIndicatorsReport::ENABLE_RIGHT;
+      },
+      1.0, 1.0);
+    auto hazard_windows = collect_signal_active_windows<HazardLightsReport>(
+      sync_data->hazard_lights_history, trajectory_start_time,
+      [](const auto & msg) { return msg.report == HazardLightsReport::ENABLE; }, 1.0, 1.0);
```
go-sakayori: Cross-reference to the comment in `lane_keeping.hpp:31-32`: these two `1.0, 1.0` literals are where the unused `lane_change_pre_grace_time` / `lane_change_post_grace_time` params should be plumbed (or where the literals stop hardcoding the grace values if the params are dropped instead).

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/lane_keeping.cpp
bskkimm: Fixed in dea7be76. I rewrote the queue progress window to select the earliest in-window sample before subtracting, and added a regression test covering the off-by-one case.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/lane_keeping.hpp
bskkimm: Fixed in dea7be76. The grace fields are now wired into both turn-indicator and hazard-light windows, exposed in the YAML config, and loaded by the node. The configured default is 1.0s to preserve the validated LK result.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/lane_keeping.hpp
bskkimm: Fixed in dea7be76. I dropped the unused LaneKeepingResult/debug fields and kept the public helper score-only for this PR.

planning/autoware_planning_data_analyzer/src/metrics/trajectory_metrics.cpp
bskkimm: Handled. The hardcoded literals were replaced with `lane_keeping_params.lane_change_pre_grace_time` and `lane_change_post_grace_time` for both signal sources.

planning/autoware_planning_data_analyzer/config/planning_data_analyzer.param.yaml
```diff
@@ -30,6 +30,7 @@
     tf_topic: "/tf"
     acceleration_topic: "/localization/acceleration"
     steering_topic: "/vehicle/status/steering_status"
+    hazard_lights_topic: "/vehicle/status/hazard_lights_status"
```
go-sakayori, line 33: memo: Check if this topic is outputted from DLR

go-sakayori (commented):
Follow-up — small style / library nits on the post-fix-up state. All non-blocking. Posting in case you do another small touch-up.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/lane_keeping.cpp
```diff
@@ -32,18 +34,54 @@ double calculate_lane_keeping_score(
...
+    const bool queue_signal_available = std::isfinite(evaluation_point.speed_mps) &&
+                                        std::isfinite(evaluation_point.cumulative_progress_m);
+    const bool queue_exempt = queue_signal_available &&
+                              evaluation_point.speed_mps <= parameters.queue_speed_threshold &&
+                              progress_window_distance <= parameters.queue_progress_threshold;
+    if (queue_exempt) {
+      queue_release_until_s = time_s + parameters.queue_release_grace_time;
+    }
+    const bool queue_release_exempt =
+      !queue_exempt && queue_release_until_s.has_value() && time_s <= *queue_release_until_s;
```
go-sakayori, line 76: Three small style nits in this block, bundled to avoid noise:

1. **Line 42 — `evaluation_points.at(index)`**: the surrounding `for (std::size_t index = 0; index < evaluation_points.size(); ++index)` guarantees the index is valid, so `evaluation_points[index]` reads cleaner. Same nit I flagged in PR #429's TLC code.
2. **Lines 44-45 — `const bool finite = std::isfinite(...); if (!finite)`**: the named bool isn't reused, so `if (!std::isfinite(evaluation_point.lateral_deviation))` would be one line shorter and just as clear.
3. **Optional dereference style is inconsistent**: this function uses `*queue_release_until_s` at line 77 (modern idiom) but `violation_start_time.value()` lower down at line 90 (older style) — both are `std::optional` in the same function. Either form is fine, but consistency aids the eye.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/lane_keeping.cpp
```diff
@@ -32,18 +34,54 @@ double calculate_lane_keeping_score(
...
+
+    if (
+      evaluation_point.is_in_intersection || lane_change_exempt || queue_exempt ||
+      queue_release_exempt || !over_threshold) {
+      reset_violation_run();
+      continue;
+    }
```
go-sakayori, line 85: Suggestion: a one-line comment above this branch naming the exemptions would make the heart of the new LK logic much easier to skim. Something like:

```cpp
// Reset the violation run on any exemption: intersection samples, signalled lane-change
// windows, queue, queue-release grace, or sub-threshold deviation.
```

The five-condition chain is the most important behavioral change in this PR and currently a future reader has to bounce between four named-bool definitions earlier in the loop to know what each flag means. Non-blocking.

planning/autoware_planning_data_analyzer/src/bag_handler.hpp
```diff
@@ -240,22 +248,44 @@ struct BagData
...
+        if (signal_buffer && synchronized_data->trajectory) {
+          const auto traj_stamp_ns =
+            rclcpp::Time(synchronized_data->trajectory->header.stamp).nanoseconds();
+          status = signal_buffer->get_closest(traj_stamp_ns, tolerance_ms);
+          const auto history_start_ns =
+            traj_stamp_ns - static_cast<rcutils_time_point_value_t>(2.0e9);
+          const auto history_end_ns =
+            traj_stamp_ns + static_cast<rcutils_time_point_value_t>(
+                              (trajectory_evaluation_horizon_s + 2.0) * 1.0e9);
+          history = signal_buffer->get_range(history_start_ns, history_end_ns);
```
go-sakayori: Style nit — `2.0e9` is "2 seconds in nanoseconds." Worth promoting to a named `constexpr` at file scope (e.g. `constexpr rcutils_time_point_value_t kSignalHistoryPaddingNs = 2'000'000'000;`) and using it for both `history_start_ns` and the additive term in `history_end_ns`. As written, a future reader has to mentally decode the literal twice.

Longer-term thought (not for this PR): this 2s padding is conceptually related to the new `lane_change_pre_grace_time` / `lane_change_post_grace_time` params — the history window needs to extend at least as far as the LK grace expects to see signal transitions. Right now they're independent constants; eventually deriving the bag-handler padding from `max(pre, post) + epsilon` would keep them aligned automatically.

planning/autoware_planning_data_analyzer/src/metrics/trajectory_metrics.cpp
```diff
@@ -46,6 +48,66 @@ using autoware::route_handler::RouteHandler;
...
+  std::vector<std::pair<double, double>> merged;
+  merged.push_back(windows.front());
+  for (std::size_t index = 1; index < windows.size(); ++index) {
+    auto & current = merged.back();
+    if (windows.at(index).first <= current.second) {
+      current.second = std::max(current.second, windows.at(index).second);
+    } else {
+      merged.push_back(windows.at(index));
+    }
+  }
```
go-sakayori, line 108: Tiny style nit — `windows.at(index)` is called twice inside the loop body. Binding once is slightly cleaner:

```cpp
for (std::size_t index = 1; index < windows.size(); ++index) {
  auto & current = merged.back();
  const auto & next_window = windows.at(index);
  if (next_window.first <= current.second) {
    current.second = std::max(current.second, next_window.second);
  } else {
    merged.push_back(next_window);
  }
}
```

(Also note: `current` is a reference into `merged`, and `merged.push_back(...)` may reallocate and invalidate it — that's currently safe because `push_back` is the last statement in its branch and `current` is rebound at the top of the next iteration. Just worth being aware of when touching this loop.)

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/lane_keeping.cpp
bskkimm: Applied. Replaced the indexed access with [], inlined the finite check, and made optional dereference style consistent.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/lane_keeping.cpp
bskkimm: Applied. Added the short exemption comment above the reset branch.

planning/autoware_planning_data_analyzer/src/bag_handler.hpp
bskkimm: Applied. Promoted the 2s signal-history padding to a named constexpr and reused it for both bounds.

planning/autoware_planning_data_analyzer/src/metrics/trajectory_metrics.cpp
bskkimm: Applied. Bound `windows.at(index)` once as `next_window` before the merge comparison.

planning/autoware_planning_data_analyzer/config/planning_data_analyzer.param.yaml
bskkimm: Acknowledged. I left this as a memo for now because it is a topic-contract question outside the LK logic patch.

go-sakayori (approved):

PR #434, opened 2026-05-21
feat(planning_data_analyzer): migrate HC metric

bskkimm:
## DES
This is the HC score PR in the EPDMS patch series.

## Scope
Migrate HC to the NAVSIM-style history-comfort calculation. This PR does not add HC debug visualization.

## Logic change
As-is:
- Planned trajectory only.
- Comfort signals were derived from future trajectory finite differences.

To-be:
- Build one padded HC sequence: recent ego odometry/acceleration history + planned 4 s trajectory.
- Reuse the shared comfort-signal helper.
- Apply NAVSIM HC thresholds to longitudinal acceleration, lateral acceleration, jerk magnitude, longitudinal jerk, yaw rate, and yaw acceleration.
- Publish HC availability/reason outputs consistently with the other migrated EPDMS scores.

## Validation
- HC matched the reference baseline: `66` non-1, `0` unavailable, reason counts: `available: 8695`

### DIFF
--- a/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/history_comfort.cpp
+++ b/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/history_comfort.cpp
@@ -14,12 +14,13 @@
 
 #include "history_comfort.hpp"
 
+#include "metrics/geometry/comfort_signal.hpp"
 #include "metrics/geometry/metric_utils.hpp"
-
-#include <autoware_utils_math/normalization.hpp>
 
 #include <algorithm>
 #include <cmath>
+#include <limits>
+#include <memory>
 #include <vector>
 
 namespace autoware::planning_data_analyzer::metrics
@@ -28,20 +29,95 @@ namespace autoware::planning_data_analyzer::metrics
 namespace
 {
 
-double calculate_time_resolution(
-  const autoware_planning_msgs::msg::TrajectoryPoint & point,
-  const autoware_planning_msgs::msg::TrajectoryPoint & next_point, const double epsilon)
+struct ComfortState
 {
-  const double time_diff = rclcpp::Duration(next_point.time_from_start).seconds() -
-                           rclcpp::Duration(point.time_from_start).seconds();
-  return time_diff > epsilon ? time_diff : epsilon;
+  double time_s{0.0};
+  geometry_msgs::msg::Pose pose;
+  double longitudinal_acceleration_mps2{0.0};
+  double lateral_acceleration_mps2{0.0};
+};
+
+double stamp_to_seconds(const rclcpp::Time & stamp)
+{
+  return static_cast<double>(stamp.nanoseconds()) * 1.0e-9;
 }
 
-void backfill_last_value_from_previous(std::vector<double> & values)
+template <typename MessageT>
+std::shared_ptr<const MessageT> closest_message(
+  const std::vector<std::shared_ptr<const MessageT>> & history, const rclcpp::Time & target,
+  const double tolerance_s)
 {
-  if (values.size() > 1U) {
-    values.back() = values[values.size() - 2];
+  std::shared_ptr<const MessageT> best;
+  double best_diff_s = std::numeric_limits<double>::max();
+  for (const auto & msg : history) {
+    if (!msg) {
+      continue;
+    }
+    const double diff_s =
+      std::abs(stamp_to_seconds(rclcpp::Time(msg->header.stamp)) - stamp_to_seconds(target));
+    if (diff_s < best_diff_s) {
+      best_diff_s = diff_s;
+      best = msg;
+    }
   }
+  return best_diff_s <= tolerance_s ? best : nullptr;
+}
+
+ComfortState make_state_from_odometry(
+  const Odometry & odometry, const AccelWithCovarianceStamped * acceleration,
+  const double relative_time_s)
+{
+  ComfortState state;
+  state.time_s = relative_time_s;
+  state.pose = odometry.pose.pose;
+  if (acceleration) {
+    state.longitudinal_acceleration_mps2 = acceleration->accel.accel.linear.x;
+    state.lateral_acceleration_mps2 = acceleration->accel.accel.linear.y;
+  }
+  return state;
+}
+
+ComfortState make_state_from_trajectory_point(
+  const autoware_planning_msgs::msg::TrajectoryPoint & point, const double relative_time_s)
+{
+  ComfortState state;
+  state.time_s = relative_time_s;
+  state.pose = point.pose;
+  state.longitudinal_acceleration_mps2 = point.acceleration_mps2;
+  return state;
+}
+
+std::vector<ComfortState> build_padded_states(
+  const SynchronizedData & sync_data, const HistoryComfortParameters & parameters)
+{
+  std::vector<ComfortState> states;
+  if (!sync_data.trajectory || parameters.sample_interval_s <= 0.0) {
+    return states;
+  }
+
+  const auto trajectory_start = rclcpp::Time(sync_data.trajectory->header.stamp);
+  const double dt = parameters.sample_interval_s;
+  for (double t = -parameters.past_horizon_s; t < -dt; t += dt) {
+    const auto target = trajectory_start + rclcpp::Duration::from_seconds(t);
+    const auto odom = closest_message(sync_data.kinematic_state_history, target, dt * 0.75);
+    if (!odom) {
+      continue;
+    }
+    const auto accel = closest_message(sync_data.acceleration_history, target, dt * 0.75);
+    states.push_back(make_state_from_odometry(*odom, accel.get(), t));
+  }
+
+  for (const auto & point : sync_data.trajectory->points) {
+    const double t = rclcpp::Duration(point.time_from_start).seconds();
+    if (t < 0.0) {
+      continue;
+    }
+    if (t > parameters.future_horizon_s + 1.0e-6) {
+      break;
+    }
+    states.push_back(make_state_from_trajectory_point(point, t));
+  }
+  return states;
 }
 
 bool is_within(const std::vector<double> & values, const double min_value, const double max_value)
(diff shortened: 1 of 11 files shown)


planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/history_comfort.cpp
```diff
@@ -61,60 +137,35 @@ bool is_abs_within(const std::vector<double> & values, const double max_abs_valu
...
 {
-  const size_t num_points = trajectory.points.size();
-  if (num_points == 0U) {
+  const auto states = build_padded_states(sync_data, history_comfort_params);
+  if (states.empty()) {
+    metrics.history_comfort = 1.0;
+    metrics.history_comfort_available = true;
+    metrics.history_comfort_reason = "available_navsim_default_pass";
     return;
   }
```
go-sakayori, line 165: Worth reconsidering: this sets `available=true` + `score=1.0` when no states could be built (i.e. no past motion data and no usable trajectory points). The summary aggregation at `open_loop_evaluator.cpp:1697-1707` collects `history_comfort_values` only when `available=true`, so each "default pass" pushes a 1.0 into the aggregate and biases the mean upward.

The other migrated subscores (LK, DAC, TTC) report `available=false` with an `unavailable_*` reason in this situation — "no data → can't evaluate" rather than "no data → pass by default." Two options:

1. Flip to `available=false` with reason like `unavailable_no_motion_history` for consistency with peer subscores.
2. Keep the current convention (if NAVSIM really wants HC to default-pass when inputs are missing) but add a code comment naming the convention and explaining why HC differs from LK/DAC/TTC. Otherwise a future reader will read this as an oversight and "fix" it inconsistently.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/history_comfort.cpp
```diff
@@ -28,20 +29,95 @@ namespace autoware::planning_data_analyzer::metrics
...
 {
-  const double time_diff = rclcpp::Duration(next_point.time_from_start).seconds() -
-                           rclcpp::Duration(point.time_from_start).seconds();
-  return time_diff > epsilon ? time_diff : epsilon;
+  ComfortState state;
+  state.time_s = relative_time_s;
+  state.pose = point.pose;
+  state.longitudinal_acceleration_mps2 = point.acceleration_mps2;
+  return state;
 }
```
go-sakayori, line 102: FYI / suggest a code comment: future trajectory samples leave `lateral_acceleration_mps2` at its default 0, while past samples (in `make_state_from_odometry` above) populate it from `accel.accel.linear.y` (IMU/EKF). After `compute_comfort_signals` smooths the input, the resulting lateral-acceleration array is effectively IMU-from-past + zeros-in-future.

The old code derived future lateral_accel from `longitudinal_velocity * yaw_rate` (centripetal). Removing that is the intended NAVSIM history-focused behavior (matches the deleted test `HistoryComfortUsesYawRateForLateralAccelerationWhenLateralVelocityIsZero`), but it's not obvious from reading this function alone. A one-line comment like:

```cpp
// Trajectory points don't carry lateral acceleration; NAVSIM HC is history-focused, so future
// lateral_acceleration_mps2 is left at 0 by design rather than derived from yaw rate.
```

…would prevent the next reader from "fixing" this by reintroducing the velocity*yaw_rate computation.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/history_comfort.cpp
```diff
@@ -28,20 +29,95 @@ namespace autoware::planning_data_analyzer::metrics
...
-  if (values.size() > 1U) {
-    values.back() = values[values.size() - 2];
+  std::vector<ComfortState> states;
+  if (!sync_data.trajectory || parameters.sample_interval_s <= 0.0) {
+    return states;
+  }
+
+  const auto trajectory_start = rclcpp::Time(sync_data.trajectory->header.stamp);
+  const double dt = parameters.sample_interval_s;
+  for (double t = -parameters.past_horizon_s; t < -dt; t += dt) {
```
go-sakayori: Past-loop boundary looks off — please confirm intent.

With `past_horizon_s = 1.5` and `dt = 0.1`, the condition `t < -dt` (i.e. `t < -0.1`) stops the loop at `t = -0.2`. So the sample at `t = -0.1` is skipped. The future loop then starts at `t = 0` (the trajectory's first point). That leaves a 2 × `dt` gap (-0.2 → 0) instead of the more natural 1 × `dt` gap (-0.1 → 0).

If the intent was to avoid double-sampling at the boundary (since the trajectory's t=0 point covers it), `t < 0.0` would still skip t=0 from the past loop and include t=-0.1. As written, you lose one past sample.

If the extra-gap behavior is intentional (e.g. to leave headroom for the polynomial filter), a one-line comment naming it would help. Otherwise: `t < 0.0` (or `t < -dt + dt/2` if you want FP-accumulation safety) looks correct.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/history_comfort.cpp
```diff
@@ -28,20 +29,95 @@ namespace autoware::planning_data_analyzer::metrics
...
+{
+  ComfortState state;
+  state.time_s = relative_time_s;
+  state.pose = odometry.pose.pose;
+  if (acceleration) {
+    state.longitudinal_acceleration_mps2 = acceleration->accel.accel.linear.x;
+    state.lateral_acceleration_mps2 = acceleration->accel.accel.linear.y;
+  }
+  return state;
+}
```
go-sakayori, line 90: FYI — this assumes `AccelWithCovarianceStamped::accel.accel.linear.x/y` is in the vehicle body frame (`base_link`). Production EKF output usually is, but the message's `header.frame_id` is the source of truth and isn't checked here.

The old HC code used `point.longitudinal_velocity_mps`, which is body-frame by trajectory convention, so the frame question didn't arise. Now if the upstream acceleration topic is ever republished in `map` or `odom` frame, the values would be in world frame and need rotation by the current yaw.

Not blocking — but a one-line comment naming the assumption (or a `RCLCPP_WARN_ONCE` if `header.frame_id != "base_link"`) would catch this if it ever changes.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/history_comfort.cpp
```diff
@@ -28,20 +29,95 @@ namespace autoware::planning_data_analyzer::metrics
...
+    }
+    const double diff_s =
+      std::abs(stamp_to_seconds(rclcpp::Time(msg->header.stamp)) - stamp_to_seconds(target));
+    if (diff_s < best_diff_s) {
+      best_diff_s = diff_s;
+      best = msg;
+    }
+  }
+  return best_diff_s <= tolerance_s ? best : nullptr;
+}
```
go-sakayori, line 75: Style / perf nits in this block — non-blocking:

1. **O(n²) lookup**: `closest_message` linearly scans the entire history per query, and `build_padded_states` calls it twice per past sample (~30 calls × ~100 messages ≈ 3000 ops per HC call). The histories are time-sorted by `Buffer::get_range`, so `std::lower_bound` could make this O(n_query × log n_history). Not catastrophic, but cheaper.
2. **`stamp_to_seconds` duplicates `rclcpp::Time::seconds()`** — `rclcpp::Time(stamp).seconds()` is the same call. The wrapper adds a hop without a clear win.
3. **`dt * 0.75` tolerance at lines 102, 106** — magic number. Either name it (`constexpr double kClosestSampleToleranceFactor = 0.75;`) with a one-line rationale, or use `dt * 0.5` which has the obvious "half-interval, no overlap" semantic.

go-sakayori (commented):
Reviewed the HC migration. Code shape is clean — the old in-line finite-difference math is gone, the padded sequence is built explicitly, and the heavy lifting delegates to the existing `compute_comfort_signals` helper. Baseline match (66 non-1) is good evidence the score is stable.

Five findings inline, in roughly descending order of importance. The first three are real questions; the last two are FYIs / nits.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/history_comfort.cpp
bskkimm: Addressed. Flipped to available=false with reason unavailable_no_motion_history and score=0.0 for consistency with peer metrics.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/history_comfort.cpp
bskkimm: Addressed. Added a comment clarifying that future lateral acceleration is left at 0.0 by design to prevent re-derivation from yaw rate.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/history_comfort.cpp
bskkimm: Addressed. Updated the loop condition to t < -dt / 2.0 to include the t = -dt sample and close the gap.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/history_comfort.cpp
bskkimm: Addressed. Added a comment noting the assumption that acceleration data is in the base_link frame.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/history_comfort.cpp
bskkimm: Addressed. Optimized `closest_message` with `std::lower_bound`, removed redundant `stamp_to_seconds`, and introduced kClosestSampleToleranceFactor.

go-sakayori (approved):

PR #435, opened 2026-05-21
feat(planning_data_analyzer): migrate EC metric

bskkimm:
## Description
This is the EC score PR in the EPDMS patch series.

## Scope
Migrate EC to the NAVSIM-style two-frame extended-comfort calculation. This PR does not add EC debug visualization.

## Logic change
As-is pipeline:
previous trajectory + current trajectory -> compare whole finite-difference signal sequences -> threshold EC

To-be pipeline:
previous/current trajectory stamps -> observation interval -> shifted overlapping horizon -> shared comfort-signal extraction -> RMS checks for acceleration magnitude, jerk magnitude, yaw rate, and yaw acceleration -> EC

## Validation
- Takanawa cumulative run through EC: `/home/beomseokkim2/rosbag/x2_takanawa/run_logs/20260521_160201`
- Evaluated trajectories: `8695`
- EC matched the reference baseline: `351` non-1, `1` unavailable, changed timestamps `0`
- DAC/DDC/TLC/TTC/LK/HC stayed matched. NC differs only by the accepted PR #426 de-dup fix.

### DIFF
--- a/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/extended_comfort.cpp
+++ b/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/extended_comfort.cpp
@@ -14,9 +14,9 @@
 
 #include "extended_comfort.hpp"
 
-#include "metrics/geometry/metric_utils.hpp"
+#include "metrics/geometry/comfort_signal.hpp"
 
-#include <autoware_utils_math/normalization.hpp>
+#include <rclcpp/rclcpp.hpp>
 
 #include <algorithm>
 #include <cmath>
@@ -28,13 +28,7 @@ namespace autoware::planning_data_analyzer::metrics
 namespace
 {
 
-struct DynamicSignals
-{
-  std::vector<double> accelerations;
-  std::vector<double> jerks;
-  std::vector<double> yaw_rates;
-  std::vector<double> yaw_accelerations;
-};
+constexpr double kDefaultTrajectoryDt = 0.1;
 
 double get_time_seconds(const builtin_interfaces::msg::Duration & time_from_start)
 {
@@ -42,69 +36,44 @@ double get_time_seconds(const builtin_interfaces::msg::Duration & time_from_star
          static_cast<double>(time_from_start.nanosec) * 1e-9;
 }
 
-double compute_rms_difference(const std::vector<double> & lhs, const std::vector<double> & rhs)
+double trajectory_dt_s(const autoware_planning_msgs::msg::Trajectory & trajectory)
 {
-  const auto count = std::min(lhs.size(), rhs.size());
+  if (trajectory.points.size() < 2U) {
+    return kDefaultTrajectoryDt;
+  }
+  const double dt = get_time_seconds(trajectory.points.at(1).time_from_start) -
+                    get_time_seconds(trajectory.points.front().time_from_start);
+  return dt > 0.0 ? dt : kDefaultTrajectoryDt;
+}
+
+std::vector<ComfortSignalInput> make_overlap_signal_inputs(
+  const autoware_planning_msgs::msg::Trajectory & trajectory, const std::size_t start_index,
+  const std::size_t count, const double dt)
+{
+  std::vector<ComfortSignalInput> inputs;
+  inputs.reserve(count);
+  for (std::size_t index = 0; index < count; ++index) {
+    const auto & point = trajectory.points.at(start_index + index);
+    inputs.push_back(
+      ComfortSignalInput{
+        static_cast<double>(index) * dt, point.pose, point.acceleration_mps2, 0.0});
+  }
+  return inputs;
+}
+
+double rms_difference(const std::vector<double> & current, const std::vector<double> & previous)
+{
+  const auto count = std::min(current.size(), previous.size());
   if (count == 0U) {
     return 0.0;
   }
 
-  double sum_squared_error = 0.0;
-  for (std::size_t i = 0; i < count; ++i) {
-    const double error = lhs.at(i) - rhs.at(i);
-    sum_squared_error += error * error;
+  double sum_squared = 0.0;
+  for (std::size_t index = 0; index < count; ++index) {
+    const double delta = current.at(index) - previous.at(index);
+    sum_squared += delta * delta;
   }
-
-  return std::sqrt(sum_squared_error / static_cast<double>(count));
-}
-
-DynamicSignals calculate_dynamic_signals(
-  const autoware_planning_msgs::msg::Trajectory & trajectory, const double epsilon)
-{
-  DynamicSignals signals;
-  const auto num_points = trajectory.points.size();
-  if (num_points < 2U) {
-    return signals;
-  }
-
-  signals.accelerations.resize(num_points - 1U, 0.0);
-  signals.yaw_rates.resize(num_points - 1U, 0.0);
-
-  for (std::size_t i = 0; i + 1U < num_points; ++i) {
-    const auto & point = trajectory.points.at(i);
-    const auto & next_point = trajectory.points.at(i + 1U);
-    const double dt = std::max(
-      get_time_seconds(next_point.time_from_start) - get_time_seconds(point.time_from_start),
-      epsilon);
-
-    const double speed = std::hypot(point.longitudinal_velocity_mps, point.lateral_velocity_mps);
-    const double next_speed =
-      std::hypot(next_point.longitudinal_velocity_mps, next_point.lateral_velocity_mps);
-    signals.accelerations.at(i) = (next_speed - speed) / dt;
-
-    const double yaw = get_yaw(point.pose.orientation);
-    const double next_yaw = get_yaw(next_point.pose.orientation);
-    const double delta_yaw = autoware_utils_math::normalize_radian(next_yaw - yaw);
-    signals.yaw_rates.at(i) = delta_yaw / dt;
-  }
-
-  if (signals.accelerations.size() >= 2U) {
-    signals.jerks.resize(signals.accelerations.size() - 1U, 0.0);
-    signals.yaw_accelerations.resize(signals.yaw_rates.size() - 1U, 0.0);
-
-    for (std::size_t i = 0; i + 1U < signals.accelerations.size(); ++i) {
-      const auto & point = trajectory.points.at(i);
-      const auto & next_point = trajectory.points.at(i + 1U);
-      const double dt = std::max(
-        get_time_seconds(next_point.time_from_start) - get_time_seconds(point.time_from_start),
-        epsilon);
-      signals.jerks.at(i) = (signals.accelerations.at(i + 1U) - signals.accelerations.at(i)) / dt;
-      signals.yaw_accelerations.at(i) =
-        (signals.yaw_rates.at(i + 1U) - signals.yaw_rates.at(i)) / dt;
-    }
-  }
-
-  return signals;
+  return std::sqrt(sum_squared / static_cast<double>(count));
 }
 
 bool are_parameters_valid(const ExtendedComfortParameters & parameters)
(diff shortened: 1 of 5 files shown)


planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/extended_comfort.cpp
```diff
@@ -121,43 +90,60 @@ ExtendedComfortResult calculate_extended_comfort(
   const autoware_planning_msgs::msg::Trajectory & current_trajectory,
   const ExtendedComfortParameters & parameters)
 {
+  ExtendedComfortResult result;
   if (!are_parameters_valid(parameters)) {
```
go-sakayori, line 94: `parameters.finite_difference_epsilon` is dead plumbing now.

It's still declared in `extended_comfort.hpp:31` (default 1e-3), loaded from YAML at `config/planning_data_analyzer.param.yaml:76`, wired into params at `autoware_planning_data_analyzer_node.cpp:159-160`, and validated as `> 0.0` inside `are_parameters_valid` (called here). But the old denominator-floor usage is gone — `grep` finds zero other references. Net effect: editing the YAML key does nothing, and a user who legitimately sets it to 0 gets `unavailable_invalid_parameters` for no functional reason.

Same anti-pattern as PR #432 LK's `lane_change_*_grace_time` params (which were resolved by wiring them through). Two options:

1. Drop the field, the YAML key, the node-side load, and the validation clause — cleanest, since nothing in the new logic uses it.
2. Drop just the validation clause if there's a planned future use, so users aren't forced to provide a value that's ignored.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/extended_comfort.cpp
```diff
@@ -28,83 +28,52 @@ namespace autoware::planning_data_analyzer::metrics
...
+  std::vector<ComfortSignalInput> inputs;
+  inputs.reserve(count);
+  for (std::size_t index = 0; index < count; ++index) {
+    const auto & point = trajectory.points.at(start_index + index);
+    inputs.push_back(
+      ComfortSignalInput{
+        static_cast<double>(index) * dt, point.pose, point.acceleration_mps2, 0.0});
   }
+  return inputs;
+}
```
go-sakayori, line 63: Two things worth addressing in this helper.

**1. Trajectory time spacing is silently assumed uniform.** `dt` is computed once from `current_trajectory.points[1].time_from_start - points[0].time_from_start` (in `trajectory_dt_s` above) and reused as `static_cast<double>(index) * dt` to synthesize each sample's time. The downstream Savitzky-Golay filter inside `compute_comfort_signals` uses these times to fit local polynomials — if the actual `time_from_start` spacing is non-uniform, the filter's derivatives (jerk, yaw_acceleration) will be miscalibrated, and the resulting RMS comparison will silently use wrong derivative values.

The old code computed dt per segment from `time_from_start` differences, so it handled non-uniform spacing. Autoware planners typically emit uniformly-spaced trajectories, but the assumption is now silent. Suggested fix:

```cpp
ComfortSignalInput{
  get_time_seconds(point.time_from_start) - get_time_seconds(trajectory.points.at(start_index).time_from_start),
  point.pose, point.acceleration_mps2, 0.0}
```

…which forwards the actual relative time and keeps the helper correct for non-uniform inputs.

**2. `lateral_acceleration_mps2 = 0` deserves a one-line comment.** Same shape as HC's `make_state_from_trajectory_point` (which got a comment after PR #434 review): trajectory points don't carry lateral acceleration, so 0 is passed, and `acceleration_magnitudes` collapses to `|longitudinal|`. Consistent comment here would prevent the next reader from wondering why the lateral channel was dropped:

```cpp
// Trajectory points don't carry lateral acceleration; EC magnitudes therefore collapse to
// longitudinal-only, matching the prepared NAVSIM reference.
```

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/extended_comfort.cpp
```diff
@@ -121,43 +90,60 @@ ExtendedComfortResult calculate_extended_comfort(
...
 
-  if (previous_trajectory.points.size() < 3U || current_trajectory.points.size() < 3U) {
-    return {0.0, false, "unavailable_short_trajectory"};
+  const double dt = trajectory_dt_s(current_trajectory);
+  const double raw_observation_interval =
+    (rclcpp::Time(current_trajectory.header.stamp) - rclcpp::Time(previous_trajectory.header.stamp))
+      .seconds();
+  const double observation_interval =
+    raw_observation_interval > 0.0 ? raw_observation_interval : dt;
+  const auto overlap_shift = static_cast<std::size_t>(std::llround(observation_interval / dt));
```
go-sakayori, line 109: The fallback on non-positive `raw_observation_interval` silently masks reversed/clock-glitched inputs.

If `current_trajectory.header.stamp <= previous_trajectory.header.stamp`, the function pretends current is exactly `dt` ahead and proceeds to emit a `score`/`available=true` verdict. Equal-stamp inputs being absorbed as "one dt ahead" is reasonable noise-handling, but **negative** intervals (previous newer than current) almost certainly indicate a bug at the caller — bag misordering, clock reset, or trajectories swapped — and silently scoring them produces an unreliable EC value that's reported as `available=true`.

Suggested split:

```cpp
if (raw_observation_interval < 0.0) {
  result.reason = "unavailable_invalid_time_alignment";
  return result;
}
const double observation_interval =
  raw_observation_interval > 0.0 ? raw_observation_interval : dt;
```

At minimum, a `RCLCPP_WARN_ONCE` for negative intervals would help diagnose data issues without changing behavior.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/extended_comfort.cpp
```diff
@@ -121,43 +90,60 @@ ExtendedComfortResult calculate_extended_comfort(
...
+  if (dt <= 0.0 || overlap_shift == 0U) {
```
go-sakayori: Minor — the `dt <= 0.0` half of this guard is unreachable.

`trajectory_dt_s` (line 39-47) returns `kDefaultTrajectoryDt = 0.1` whenever the computed dt is non-positive, so `dt > 0.0` always holds by the time we reach this line. Either drop the `dt <= 0.0` clause (keeping just `overlap_shift == 0U`), or push the responsibility upstream by having `trajectory_dt_s` return `std::optional<double>` and propagating the failure. The cleanest is to just simplify the condition since the helper's invariant is already strong enough.

go-sakayori (commented):
Reviewed the EC migration. The shift-and-align approach (overlap_shift = round(observation_interval / dt), compare `current[0..N]` against `previous[shift..shift+N]`) reads cleanly and matches the NAVSIM intent; delegating signal extraction to the shared `compute_comfort_signals` helper keeps the file small. Baseline match (351 non-1, 1 unavailable, 0 changed timestamps) is good evidence.

Four findings inline, descending importance.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/extended_comfort.cpp
bskkimm: Removed the dead EC `finite_difference_epsilon` plumbing from the parameter struct, YAML, node loading, and validation. `HC finite_difference_epsilon` remains because HC still uses it.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/extended_comfort.cpp
bskkimm: Updated overlap signal input generation to use actual relative `time_from_start` values instead of `index * dt`, and added the requested comment explaining the longitudinal-only acceleration magnitude for planned trajectories.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/extended_comfort.cpp
bskkimm: Added an explicit negative observation-interval guard returning `unavailable_invalid_time_alignment`. Equal timestamps still fall back to one trajectory dt as before.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/extended_comfort.cpp
bskkimm: Simplified the guard by removing the unreachable dt <= 0.0 condition; trajectory_dt_s already guarantees a positive fallback dt.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/extended_comfort.cpp
```diff
@@ -14,9 +14,9 @@
 
 #include "extended_comfort.hpp"
 
-#include "metrics/geometry/metric_utils.hpp"
+#include "metrics/geometry/comfort_signal.hpp"
 
-#include <autoware_utils_math/normalization.hpp>
+#include <rclcpp/rclcpp.hpp>
```
go-sakayori: Two small library-usage nits:

1. **Narrow the rclcpp include.** Only `rclcpp::Time` and `rclcpp::Duration::from_seconds` are used in this TU, so `<rclcpp/time.hpp>` (or `<rclcpp/duration.hpp>` + `<rclcpp/time.hpp>`) would be enough — `<rclcpp/rclcpp.hpp>` is the full umbrella header and is overkill here.
2. **`get_time_seconds` helper duplicates a built-in.** `rclcpp::Duration(time_from_start).seconds()` returns the same value as the local helper at line 33-37. This exactly parallels the `stamp_to_seconds` cleanup you did in PR #434 HC (removed in favor of `rclcpp::Time::seconds()`) — same fix would apply here for consistency: drop the helper and call `rclcpp::Duration(point.time_from_start).seconds()` at the use sites (lines 44-45, 55, 58).

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/extended_comfort.cpp
```diff
@@ -28,90 +28,60 @@ namespace autoware::planning_data_analyzer::metrics
...
+  std::vector<ComfortSignalInput> inputs;
+  inputs.reserve(count);
+  const double start_time_s = get_time_seconds(trajectory.points.at(start_index).time_from_start);
+  for (std::size_t index = 0; index < count; ++index) {
+    const auto & point = trajectory.points.at(start_index + index);
+    const double time_s = get_time_seconds(point.time_from_start) - start_time_s;
+    // Trajectory points do not carry lateral acceleration; EC therefore compares
+    // longitudinal-only acceleration magnitudes for planned horizons.
+    inputs.push_back(ComfortSignalInput{time_s, point.pose, point.acceleration_mps2, 0.0});
   }
```
go-sakayori, line 61: Two minor style nits in this function:

1. **`.at(...)` vs `[...]` in bounded loops.** Lines 55, 57 (and 75 in `rms_difference`) use range-checked `.at()`, but the surrounding code already proves the indices are valid (`start_index + index < trajectory.points.size()` is guaranteed by the `overlap_shift >= previous_size` and `overlap_count` checks at the call site). `[]` reads slightly cleaner. Same nit pattern from PR #429 / PR #432.
2. **Comment placement.** The lateral-acceleration comment at lines 59-60 documents an invariant of the whole helper, not just the loop body — moving it to a function-level comment above `make_overlap_signal_inputs` (or even on the `ComfortSignalInput` field at the call site, matching where HC's equivalent comment lives in `make_state_from_trajectory_point`) would be slightly easier to find. Pure style preference.

go-sakayori (commented):
Follow-up — small style / library nits on the post-fix-up state. All non-blocking. Posting in case you do another small touch-up.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/extended_comfort.cpp
bskkimm: Addressed. Swapped umbrella header for specific rclcpp headers and switched to native seconds() API.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/extended_comfort.cpp
bskkimm: Addressed. Used [] for indexing in bounded loops and moved the lateral-acceleration note to the function-level comment.

go-sakayori (approved):

PR #436, opened 2026-05-21
feat(planning_data_analyzer): migrate EP metric

bskkimm:
This is the EP score PR in the EPDMS patch series.

## Scope
Migrate EP to the NAVSIM-style single selected trajectory progress calculation. Multi-candidate proposal-batch EP is left as a documented follow-up.

## Logic change
As-is pipeline:
selected trajectory + candidate trajectories -> selected-candidate matching -> best candidate progress ratio -> EP

To-be pipeline:
selected trajectory -> route-relative raw progress -> multiplicative NC/DAC/DDC/TLC mask -> masked denominator -> single-proposal NAVSIM-style EP

## Validation
- Takanawa cumulative run through EP: `/home/beomseokkim2/rosbag/x2_takanawa/run_logs/20260521_200556` was compared, then deleted after recording the summary.
- Evaluated trajectories: `8695`
- EP matched the reference baseline score: `0` non-1, `0` unavailable, changed timestamps `0`
- EP reason counts shifted only because of the accepted NC de-dup delta: fallback `5659 -> 5672`, ratio `3036 -> 3023`

### DIFF
--- a/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ego_progress.cpp
+++ b/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ego_progress.cpp
@@ -127,8 +127,8 @@ EgoProgressResult calculate_ego_progress(
   constexpr double kProgressDistanceThresholdM = 5.0;
   if (result.denominator_m > kProgressDistanceThresholdM) {
     // Note: Since we only have a single trajectory candidate in this implementation,
-    // the score simplifies to 1.0 (raw_progress / denominator where denominator = raw_progress * 1.0)
-    // if all multiplicative metrics passed.
+    // the score simplifies to 1.0 (raw_progress / denominator where denominator = raw_progress
+    // * 1.0) if all multiplicative metrics passed.
     result.score = std::clamp(result.raw_progress_m / result.denominator_m, 0.0, 1.0);
     result.reason = "available_single_proposal_navsim_ratio";
   } else {
--- a/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ego_progress.cpp
+++ b/planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ego_progress.cpp
@@ -20,11 +20,8 @@
 #include <autoware/lanelet2_utils/geometry.hpp>
 
 #include <algorithm>
-#include <cmath>
-#include <limits>
 #include <memory>
 #include <optional>
-#include <unordered_set>
 #include <vector>
 
 namespace autoware::planning_data_analyzer::metrics
@@ -36,19 +33,38 @@ namespace
 {
 
 std::optional<double> calculate_raw_progress_m(
-  const Trajectory & trajectory, const std::shared_ptr<RouteHandler> & route_handler)
+  const Trajectory & trajectory, const std::shared_ptr<RouteHandler> & route_handler,
+  std::vector<geometry_msgs::msg::Point> * route_reference_points,
+  const lanelet::ConstLanelets * route_relevant_lanelets)
 {
   if (trajectory.points.size() < 2U) {
     return std::nullopt;
   }
 
-  const auto route_lanelets = collect_route_relevant_lanelets(trajectory, route_handler);
+  const auto local_route_lanelets = route_relevant_lanelets
+                                      ? lanelet::ConstLanelets{}
+                                      : collect_route_relevant_lanelets(trajectory, route_handler);
+  const auto & route_lanelets =
+    route_relevant_lanelets ? *route_relevant_lanelets : local_route_lanelets;
   if (route_lanelets.empty()) {
     return std::nullopt;
   }
   const auto lanelet_map_ptr = route_handler->getLaneletMapPtr();
   if (!lanelet_map_ptr) {
     return std::nullopt;
+  }
+
+  if (route_reference_points) {
+    route_reference_points->clear();
+    for (const auto & lanelet : route_lanelets) {
+      for (const auto & point : lanelet.centerline()) {
+        geometry_msgs::msg::Point msg_point;
+        msg_point.x = point.x();
+        msg_point.y = point.y();
+        msg_point.z = point.z();
+        route_reference_points->push_back(msg_point);
+      }
+    }
   }
 
   const auto start_arc =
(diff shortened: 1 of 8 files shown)


planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ego_progress.hpp
```diff
@@ -34,12 +37,21 @@ struct EgoProgressResult
...
   double raw_progress_m{0.0};
   double best_raw_progress_m{0.0};
+  double multiplicative_mask{0.0};
+  double denominator_m{0.0};
+  geometry_msgs::msg::Point start_point;
+  geometry_msgs::msg::Point end_point;
+  std::vector<geometry_msgs::msg::Point> route_reference_points;
 };
 
 EgoProgressResult calculate_ego_progress(
```
go-sakayori, line 48: nits: The 10-parameter signature pairs each subscore value with an availability bool. This ordering is easy to get wrong at the call site. Consider bundling each pair into a small struct

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ego_progress.hpp
```diff
@@ -34,12 +37,21 @@ struct EgoProgressResult
   std::string reason{"unavailable"};
   double raw_progress_m{0.0};
   double best_raw_progress_m{0.0};
+  double multiplicative_mask{0.0};
+  double denominator_m{0.0};
+  geometry_msgs::msg::Point start_point;
+  geometry_msgs::msg::Point end_point;
+  std::vector<geometry_msgs::msg::Point> route_reference_points;
```
go-sakayori, line 45: These 3 fields are assigned in trajectory_metrics.cpp, but the are not consumed afterwards.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ego_progress.cpp
```diff
@@ -138,68 +102,39 @@ EgoProgressResult calculate_ego_progress(
...
+  result.best_raw_progress_m = result.denominator_m;
   result.available = true;
-  result.reason = "available";
 
   constexpr double kProgressDistanceThresholdM = 5.0;
-  if (best_raw_progress_m > kProgressDistanceThresholdM) {
-    result.score =
-      std::clamp(selected_candidate_raw_progress.value() / best_raw_progress_m, 0.0, 1.0);
+  if (result.denominator_m > kProgressDistanceThresholdM) {
+    result.score = std::clamp(result.raw_progress_m / result.denominator_m, 0.0, 1.0);
```
go-sakayori, line 132: Always clamps to 1.0, but is this intended? If so, at least please leave a comment or write it in README

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ego_progress.cpp
```diff
@@ -138,68 +102,39 @@ EgoProgressResult calculate_ego_progress(
...
     return result;
   }
-
-  result.raw_progress_m = selected_candidate_raw_progress.value();
-  result.best_raw_progress_m = best_raw_progress_m;
+  result.raw_progress_m = selected_raw_progress.value();
+  result.multiplicative_mask = no_at_fault_collision * drivable_area_compliance *
+                               driving_direction_compliance * traffic_light_compliance;
+  result.denominator_m = result.raw_progress_m * result.multiplicative_mask;
+  result.best_raw_progress_m = result.denominator_m;
```
go-sakayori: `best_raw_progress_m` is no longer a raw progress value here. Since this is still exported as `ego_progress_best_raw_m` in the full JSON, downstream users may read it as the old best raw candidate progress in meters. Can we either keep `best_raw_progress_m` semantically raw, remove/deprecate it for the single-proposal path, or stop writing the maskeddenominator into the old `*_best_raw_m` field now that `ego_progress_denominator_m` exists?

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ego_progress.cpp
```diff
@@ -51,6 +54,19 @@ std::optional<double> calculate_raw_progress_m(
...
 
+  if (route_reference_points) {
+    route_reference_points->clear();
+    for (const auto & lanelet : route_lanelets) {
+      for (const auto & point : lanelet.centerline()) {
+        geometry_msgs::msg::Point msg_point;
+        msg_point.x = point.x();
+        msg_point.y = point.y();
+        msg_point.z = point.z();
+        route_reference_points->push_back(msg_point);
```
go-sakayori:
```suggestion
  route_reference_points->push_back(
    autoware::universe_utils::createPoint(point.x(), point.y(), point.z()));
```
Also is this `route_reference_points` used?

go-sakayori (changes requested):

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ego_progress.cpp
bskkimm: Intentional for the NAVSIM-faithful single-proposal EP path. In this branch EP is saturated to 1.0 when available, so I added a short code comment to make that explicit.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ego_progress.hpp
bskkimm: Handled by bundling the multiplicative inputs into a single struct, so the call site is no longer a long value/bool chain.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ego_progress.cpp
bskkimm: `best_raw_progress_m` is restored to mean raw progress again; the masked denominator is tracked in a separate field.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ego_progress.hpp
bskkimm: These fields are kept intentionally for debug topics, so they are now consumed rather than dead state.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ego_progress.cpp
bskkimm: Applied the suggested helper. `route_reference_points` is used by the EP debug topic (`ep_route_reference`), so it is not dead state.

go-sakayori (approved):

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ego_progress.hpp
bskkimm: Addressed. Bundled subscore/availability pairs into a MetricScore struct to simplify the signature.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ego_progress.hpp
bskkimm: Addressed. These fields are now consumed by the result struct for debugging purposes.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ego_progress.cpp
bskkimm: Addressed. Added a comment clarifying that with a single trajectory candidate, the score simplifies to 1.0 when multiplicative metrics pass.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ego_progress.cpp
bskkimm: Addressed. Restored best_raw_progress_m to hold semantically raw progress; masked denominator is now handled separately.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ego_progress.cpp
bskkimm: Addressed. Switched to autoware_utils::create_point (available via existing dependencies) for point creation.

PR #437, opened 2026-05-21
feat(planning_data_analyzer): update EPDMS aggregation filter

bskkimm:
## Description
This PR updates the EPDMS aggregation path to match the NAVSIM-faithful human-filter behavior for the migrated scores.

### Changes:
- Keep EP in the human-filtered synthetic score.
- Do not human-filter EC.
- Factor the shared synthetic aggregation formulas to reduce duplication.

## Validation
- Full Takanawa run:
  - Inference time: elapsed_sec=192.75
  - Analyzer time: elapsed_sec=197.93

### DIFF
--- a/planning/autoware_planning_data_analyzer/src/metrics/epdms/aggregation/epdms_aggregation.cpp
+++ b/planning/autoware_planning_data_analyzer/src/metrics/epdms/aggregation/epdms_aggregation.cpp
@@ -23,6 +23,7 @@ namespace
 {
 
 constexpr double kHumanFilterZeroEpsilon = 1.0e-9;
+constexpr double kSyntheticWeightedDenominator = 16.0;
 
 double apply_human_filter(
   const double agent_value, const bool agent_available, const double human_value,
@@ -40,6 +41,23 @@ void populate_human_filter_metric(
   metric.human_reference_available = human_available;
   metric.filtered = apply_human_filter(
     agent_value, agent_available, human_value, human_available, metric.filter_applied);
+}
+
+double calculate_multiplicative_metrics_prod(
+  const double no_at_fault_collision, const double drivable_area_compliance,
+  const double driving_direction_compliance, const double traffic_light_compliance)
+{
+  return no_at_fault_collision * drivable_area_compliance * driving_direction_compliance *
+         traffic_light_compliance;
+}
+
+double calculate_weighted_metrics(
+  const double ego_progress, const double time_to_collision_within_bound, const double lane_keeping,
+  const double history_comfort, const double extended_comfort)
+{
+  return (5.0 * ego_progress + 5.0 * time_to_collision_within_bound + 2.0 * lane_keeping +
+          2.0 * history_comfort + 2.0 * extended_comfort) /
+         kSyntheticWeightedDenominator;
 }
 
 }  // namespace
@@ -100,7 +118,6 @@ SyntheticEpdmsMetrics calculate_synthetic_epdms(
   const EpdmsMetricSnapshot & agent_metrics, const HumanFilterMetrics & human_filter_metrics)
 {
   SyntheticEpdmsMetrics result;
-  constexpr double kSyntheticWeightedDenominator = 16.0;
 
   const bool raw_multiplicative_available = agent_metrics.no_at_fault_collision_available &&
                                             agent_metrics.drivable_area_compliance_available &&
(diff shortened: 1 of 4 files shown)


planning/autoware_planning_data_analyzer/src/metrics/epdms/aggregation/epdms_aggregation.cpp
```diff
@@ -53,11 +71,6 @@ HumanFilterMetrics calculate_human_filter_metrics(
     result.history_comfort, agent_metrics.history_comfort, agent_metrics.history_comfort_available,
     human_metrics.history_comfort, human_metrics.history_comfort_available);
 
-  populate_human_filter_metric(
-    result.extended_comfort, agent_metrics.extended_comfort,
-    agent_metrics.extended_comfort_available, human_metrics.extended_comfort,
-    human_metrics.extended_comfort_available);
-
   populate_human_filter_metric(
```
go-sakayori, line 74: `HumanFilterMetrics::extended_comfort` is never populated after this PR (`populate_human_filter_metric` for EC was removed), so the field stays at its zero-initialized default. Any future code that reads `human_filter_metrics.extended_comfort.filtered` will silently get `0.0`. Since `calculate_synthetic_epdms` now uses `agent_metrics.extended_comfort` directly, consider removing `extended_comfort` from `HumanFilterMetrics` entirely to prevent that silent footgun.

planning/autoware_planning_data_analyzer/src/metrics/epdms/aggregation/epdms_aggregation.cpp
```diff
@@ -137,20 +147,16 @@ SyntheticEpdmsMetrics calculate_synthetic_epdms(
...
+    result.human_filtered.multiplicative_metrics_prod = calculate_multiplicative_metrics_prod(
+      human_filter_metrics.no_at_fault_collision.filtered,
+      human_filter_metrics.drivable_area_compliance.filtered,
+      human_filter_metrics.driving_direction_compliance.filtered,
+      human_filter_metrics.traffic_light_compliance.filtered);
+    result.human_filtered.weighted_metrics = calculate_weighted_metrics(
+      human_filter_metrics.ego_progress.filtered,
+      human_filter_metrics.time_to_collision_within_bound.filtered,
+      human_filter_metrics.lane_keeping.filtered, human_filter_metrics.history_comfort.filtered,
+      agent_metrics.extended_comfort);
```
go-sakayori, line 160: All other arguments come from `human_filter_metrics.xxx.filtered`, but `extended_comfort` is deliberately passed as the raw `agent_metrics.extended_comfort`. A future reader could easily treat this as a bug and "fix" it to `human_filter_metrics.extended_comfort.filtered` — which is always `0.0` after this PR. A short inline comment such as `// EC bypasses human filter by design` would prevent that.

planning/autoware_planning_data_analyzer/test/metrics/test_epdms_aggregation.cpp
```diff
@@ -73,8 +90,48 @@ TEST(EpdmsAggregationTest, SyntheticEpdmsUsesFirstStageMetricGroups)
...
+  EpdmsMetricSnapshot agent;
+  agent.history_comfort = 0.5;
+  agent.history_comfort_available = true;
+  agent.ego_progress = 0.0;
+  agent.ego_progress_available = true;
+  agent.time_to_collision_within_bound = 1.0;
+  agent.time_to_collision_within_bound_available = true;
+  agent.lane_keeping = 1.0;
+  agent.lane_keeping_available = true;
+  agent.extended_comfort = 0.0;
```
go-sakayori: `agent.extended_comfort = 0.0` here means the expected score (`0.8125`) is identical regardless of whether raw or filtered EC is used — both are `0.0`. The test name promises that raw EC is used, but the values don't actually prove it. A stronger scenario would set `agent.extended_comfort` to something non-zero (e.g. `0.5`) with `human.extended_comfort = 0.0` (which would have triggered the old filter) to verify the raw value is used unambiguously.

planning/autoware_planning_data_analyzer/src/metrics/epdms/aggregation/epdms_aggregation.cpp
bskkimm: Removed `HumanFilterMetrics::extended_comfort` since EC bypasses human filtering by design in the NAVSIM-faithful synthetic score.

planning/autoware_planning_data_analyzer/src/metrics/epdms/aggregation/epdms_aggregation.cpp
bskkimm: Added an inline comment to make the EC bypass explicit, since raw agent EC is used intentionally here.

planning/autoware_planning_data_analyzer/test/metrics/test_epdms_aggregation.cpp
bskkimm: Strengthened the test so raw EC is non-zero and differs from the filtered value, which proves the bypass behavior.

go-sakayori (approved):

PR #441, opened 2026-05-27
feat(planning_data_analyzer): add EPDMS debug topics

bskkimm:
## Description

This PR adds optional EPDMS debug outputs for the migrated Autoware EPDMS subscores.

As-is:
- EPDMS subscores are emitted as scalar/array metrics.
- Detailed violation geometry is not available in the evaluation bag.

To-be:
- When `open_loop.debug_topics_enabled:=true`, the analyzer writes subscore-specific `/debug/epdms/*` topics for Lichtblick diagnosis.
- The debug topics expose violation summaries, ego/object footprints, drivable-area evidence, route/oncoming/intersection geometry, stop lines, comfort samples, progress references, and 4 s planned/GT horizon footprints.
- The default behavior remains debug-off unless explicitly enabled.

## Related links

None.

## How was this PR tested?

- Route: full Takanawa route, about 15 min.
- Device: Intel Core i9-14900K.
- Underlay used for full-run validation: https://github.com/tier4/pilot-auto.x2/tree/feat/v4.3/e2e
- Analyzer runtime: `197 s` without debug topics vs `212 s` with debug topics. The debug output added about `15 s` on this run.

## Notes for reviewers

- This PR adds debug outputs only. It does not add new subscribed inputs.
- Existing metric score topics and metric JSON fields are unchanged.
- The added outputs are intended for Lichtblick/manual diagnosis and are not consumed by the scoring pipeline.
- The debug topics are written only when the existing parameter `open_loop.debug_topics_enabled` is `true`.

#### Additions and removals

| Change type | Topic Type | Topic Name | Message Type | Description |
|:------------|:-----------|:-----------|:-------------|:------------|
| Added | Pub | `/debug/epdms/trajectory/planned_horizon_4s` | `visualization_msgs/msg/MarkerArray` | Planned 4 s ego footprint horizon. |
| Added | Pub | `/debug/epdms/trajectory/gt_horizon_4s` | `visualization_msgs/msg/MarkerArray` | Ground-truth 4 s ego footprint horizon. |
| Added | Pub | `/debug/epdms/nc/collision_summary` | `std_msgs/msg/String` | NC collision summary for the evaluated trajectory. |
| Added | Pub | `/debug/epdms/nc/ego_footprints` | `visualization_msgs/msg/MarkerArray` | NC ego footprint horizon. |
| Added | Pub | `/debug/epdms/nc/object_footprints` | `visualization_msgs/msg/MarkerArray` | NC object footprint horizon. |
| Added | Pub | `/debug/epdms/nc/overlap_areas` | `visualization_msgs/msg/MarkerArray` | NC ego/object overlap areas. |
| Added | Pub | `/debug/epdms/dac/violation_summary` | `std_msgs/msg/String` | DAC violation summary. |
| Added | Pub | `/debug/epdms/dac/ego_footprints` | `visualization_msgs/msg/MarkerArray` | DAC ego footprints and failing samples. |
| Added | Pub | `/debug/epdms/dac/admissible_*` | `visualization_msgs/msg/MarkerArray` | DAC admissible road, shoulder, intersection, hatched-road-marking, and parking areas. |
| Added | Pub | `/debug/epdms/dac/road_border_*` | `visualization_msgs/msg/MarkerArray` | DAC road-border fallback diagnostic markers. |
| Added | Pub | `/debug/epdms/dac/failing_corners` | `visualization_msgs/msg/MarkerArray` | DAC ego footprint corners outside accepted drivable area. |
| Added | Pub | `/debug/epdms/ddc/violation_summary` | `std_msgs/msg/String` | DDC violation summary. |
| Added | Pub | `/debug/epdms/ddc/ego_centers` | `visualization_msgs/msg/MarkerArray` | DDC planned ego center path. |
| Added | Pub | `/debug/epdms/ddc/oncoming_segments` | `visualization_msgs/msg/MarkerArray` | DDC path segments counted as oncoming/wrong-direction progress. |
| Added | Pub | `/debug/epdms/ddc/route_lane_polygons` | `visualization_msgs/msg/MarkerArray` | DDC route-direction lane polygons. |
| Added | Pub | `/debug/epdms/ddc/intersection_lane_polygons` | `visualization_msgs/msg/MarkerArray` | DDC intersection polygons. |
| Added | Pub | `/debug/epdms/tlc/violation_summary` | `std_msgs/msg/String` | TLC violation summary. |
| Added | Pub | `/debug/epdms/tlc/ego_footprints` | `visualization_msgs/msg/MarkerArray` | TLC ego footprints over the checked horizon. |
| Added | Pub | `/debug/epdms/tlc/stop_lines` | `visualization_msgs/msg/MarkerArray` | TLC relevant stop lines. |
| Added | Pub | `/debug/epdms/ttc/violation_summary` | `std_msgs/msg/String` | TTC violation summary. |
| Added | Pub | `/debug/epdms/ttc/ego_footprints` | `visualization_msgs/msg/MarkerArray` | TTC ego footprint sequence. |
| Added | Pub | `/debug/epdms/ttc/object_footprints` | `visualization_msgs/msg/MarkerArray` | TTC object footprint sequence. |
| Added | Pub | `/debug/epdms/ttc/overlap_areas` | `visualization_msgs/msg/MarkerArray` | TTC ego/object overlap areas. |
| Added | Pub | `/debug/epdms/lk/violation_summary` | `std_msgs/msg/String` | LK violation summary. |
| Added | Pub | `/debug/epdms/lk/ego_center_path` | `visualization_msgs/msg/MarkerArray` | LK ego center path with violation status. |
| Added | Pub | `/debug/epdms/lk/reference_centerlines` | `visualization_msgs/msg/MarkerArray` | LK reference lane centerlines. |
| Added | Pub | `/debug/epdms/hc/*` | `std_msgs/msg/String`, `tier4_metric_msgs/msg/MetricArray`, `visualization_msgs/msg/MarkerArray` | HC component status, sample times, segments, and comfort footprint markers. |
| Added | Pub | `/debug/epdms/ec/*` | `std_msgs/msg/String`, `tier4_metric_msgs/msg/MetricArray` | EC comparison summary and dynamic-signal deltas. |
| Added | Pub | `/debug/epdms/ep/*` | `std_msgs/msg/String`, `visualization_msgs/msg/MarkerArray` | EP progress summary and route progress markers. |

#### Additions and removals

None.

This PR uses the existing `open_loop.debug_topics_enabled` parameter. When it is `false`, no `/debug/epdms/*` topics are written.

## Effects on system behavior

- Default behavior is unchanged because `open_loop.debug_topics_enabled` defaults to `false`.
- When debug topics are enabled, the analyzer writes additional marker/string/array topics to the output evaluation bag for Lichtblick inspection.
- The scoring pipeline does not consume the added debug topics, so metric computation results are not intended to change because of these outputs.
- Enabling debug topics increases analyzer output size and runtime. On the tested Takanawa run, analyzer runtime increased from `197 s` to `212 s`.


go-sakayori (commented):
Please use `<autoware_utils_visualization/marker_helper.hpp>` for vizualization

go-sakayori (commented):
`autoware_utils_geometry` has `to_msg(point.to_3d(z))` so please use it

go-sakayori (commented):
please use `nlohmann::json` instead of building JSON via `ostringstream`

go-sakayori (commented):
Sorry I didn't inline the comments. Made mistake with the gh command

xtk8532704:
> Takanawa full run was validated on the pilot-X2-compatible counterpart branch with the same debug logic:
> Run: /home/beomseokkim2/rosbag/x2_takanawa/run_logs/20260527_105909
> Inference: 194.36s
> Analyzer: 212.23s
> Evaluated trajectories: 8695
> DAC/DDC/TLC/TTC/LK/HC/EC/EP matched the baseline counts.
> ADE/FDE/AHE/FHE matched exactly.
> NC differed by the known later NC behavior shift: 74 -> 87 non-1.
> Local upstream-style build was attempted, but this machine's available underlay lacks the newer upstream lanelet API symbols already used by upstream/main; CI should validate against the repository dependency set.

What info do you want the reviewer to check? Do the numbers you provided  (194.36, 8695,  etc.) have any impact on the review or at the system level?

Can you provide a **human-readable** description of your PR's effect instead of a meaningless AI summary? like, IF changes( subscribed input topic updates, output topics' format updates, output json format updates, etc.)

It's hard to integrate your changes to other parts of the system if you don't provide the detailed IF updates

bskkimm:
> > Takanawa full run was validated on the pilot-X2-compatible counterpart branch with the same debug logic:
> > Run: /home/beomseokkim2/rosbag/x2_takanawa/run_logs/20260527_105909
> > Inference: 194.36s
> > Analyzer: 212.23s
> > Evaluated trajectories: 8695
> > DAC/DDC/TLC/TTC/LK/HC/EC/EP matched the baseline counts.
> > ADE/FDE/AHE/FHE matched exactly.
> > NC differed by the known later NC behavior shift: 74 -> 87 non-1.
> > Local upstream-style build was attempted, but this machine's available underlay lacks the newer upstream lanelet API symbols already used by upstream/main; CI should validate against the repository dependency set.
> 
> What info do you want the reviewer to check? Do the numbers you provided (194.36, 8695, etc.) have any impact on the review or at the system level?
> 
> Can you provide a **human-readable** description of your PR's effect instead of a meaningless AI summary? like, IF changes( subscribed input topic updates, output topics' format updates, output json format updates, etc.)
> 
> It's hard to integrate your changes to other parts of the system if you don't provide the detailed IF updates

Understood. I will make sure to add that info from now on!

go-sakayori:
@bskkimm 
As @xtk8532704  mentioned it is import to write the interface/parameter changes. I would like you to update the PR description, referring to the style that is widely used in AWF. These are examples https://github.com/autowarefoundation/autoware_universe/pull/12653 https://github.com/autowarefoundation/autoware_universe/pull/12655

bskkimm:
> @bskkimm As @xtk8532704 mentioned it is import to write the interface/parameter changes. I would like you to update the PR description, referring to the style that is widely used in AWF. These are examples [autowarefoundation/autoware_universe#12653](https://github.com/autowarefoundation/autoware_universe/pull/12653) [autowarefoundation/autoware_universe#12655](https://github.com/autowarefoundation/autoware_universe/pull/12655)

Prameter change section added. The new parameters (topics) are all necessary for sub-score level debuggin in lichtblick.

planning/autoware_planning_data_analyzer/src/metrics/epdms/debug/epdms_debug_writer.cpp
```diff
@@ -0,0 +1,1086 @@
...
+  return debug_topic("trajectory", name);
+}
+
+std_msgs::msg::ColorRGBA make_color(
+  const float red, const float green, const float blue, const float alpha = 1.0F)
+{
+  return autoware_utils_visualization::create_marker_color(red, green, blue, alpha);
+}
+
+Marker make_delete_all_marker(const rclcpp::Time & stamp)
```
go-sakayori: This builds a default marker and then sets `action = DELETEALL`.`autoware_utils_visualization::create_deleted_default_marker(stamp, ns, id)` already does exactly this. The same header is already included on line 21, so this helper can be removed and call sites can use the utility directly.

planning/autoware_planning_data_analyzer/src/metrics/epdms/debug/epdms_debug_writer.cpp
```diff
@@ -0,0 +1,1086 @@
...
+  p1.x += half_extent;
+  p1.y -= half_extent;
+  p2.x += half_extent;
+  p2.y += half_extent;
+  p3.x -= half_extent;
+  p3.y += half_extent;
+  return {p0, p1, p2, p3, p0};
+}
+
+std::vector<geometry_msgs::msg::Point> polygon_to_msg_points(
```
go-sakayori: `metric_utils.cpp::polygon_to_points` does the same Polygon2d→`vector<geometry_msgs::Point>` conversion and forwards to `autoware_utils_geometry::to_msg`. Including `metrics/geometry/metric_utils.hpp` and dropping this local copy would deduplicate it; the trailing manual ring-close can be replaced by `boost::geometry::correct` (or by emitting the closing point unconditionally) so behavior stays the same for both ring and non-ring inputs.

planning/autoware_planning_data_analyzer/src/metrics/geometry/metric_utils.cpp
```diff
@@ -70,4 +76,60 @@ const autoware_perception_msgs::msg::PredictedPath * highest_confidence_path(
   return it == object.kinematics.predicted_paths.end() ? nullptr : &(*it);
 }
 
+std::string object_id_to_string(
```
go-sakayori, line 79: `autoware_utils_uuid::to_hex_string(unique_identifier_msgs::msg::UUID)` is the canonical hex serializer in the Autoware utils. This PR adds several new call sites for `object_id_to_string` (NC events, TTC events, horizon footprints, overlap areas), so it's a good moment to swap to the utility — the `valid` branch can be handled at the call site, or wrapped if you want to preserve the `"invalid"` sentinel.

planning/autoware_planning_data_analyzer/src/metrics/geometry/metric_utils.cpp
```diff
@@ -70,4 +76,60 @@ const autoware_perception_msgs::msg::PredictedPath * highest_confidence_path(
...
+  }
+  std::ostringstream oss;
+  oss << std::hex << std::setfill('0');
+  for (const auto byte : object_id.uuid) {
+    oss << std::setw(2) << static_cast<int>(byte);
+  }
+  return oss.str();
+}
+
+geometry_msgs::msg::Point to_msg_point(
```
go-sakayori, line 88: This is a one-line wrapper around `autoware_utils_geometry::to_msg(point.to_3d(z))`. A second copy with the same name and signature is also defined in an anonymous namespace in `trajectory_metrics.cpp` (and a third for `lanelet::ConstPoint3d` in `traffic_light_compliance.cpp`). Consider either inlining the `to_msg(point.to_3d(z))` call at use sites or keeping a single canonical helper here and using it from the other TUs.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/no_at_fault_collision.cpp
```diff
@@ -60,13 +63,40 @@ struct AtFaultCollision
...
+}
+
 double ego_speed(const autoware_planning_msgs::msg::TrajectoryPoint & point)
 {
   return std::hypot(point.longitudinal_velocity_mps, point.lateral_velocity_mps);
 }
 
-bool front_bumper_intersects(
-  const geometry_msgs::msg::Pose & ego_pose, const Polygon2d & object_polygon,
+std::vector<geometry_msgs::msg::Point> front_bumper_points(
```
go-sakayori, line 99: Both points are body-frame offsets rotated into world frame by hand using sin/cos. `autoware_utils_geometry::calc_offset_pose(ego_pose, max_longitudinal_offset_m, min_lateral_offset_m, 0.0, 0.0).position` (and the symmetric call for max_lateral) returns the same world-frame point — no need to recompute the yaw, sin/cos, or build the inline lambda.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/traffic_light_compliance.cpp
```diff
@@ -41,6 +42,57 @@ namespace
 
 using TurnDirection = autoware::experimental::lanelet2_utils::TurnDirection;
 
+geometry_msgs::msg::Point to_msg_point(const lanelet::ConstPoint3d & point, const double z)
```
go-sakayori: Duplicates `autoware_utils_geometry::to_msg(Point3d{point.x(), point.y(), z})` , and the file already depends on `autoware_utils_geometry`. The same pattern (`msg.x = ...; msg.y = ...; msg.z = z;`) also appears below at line 63 inside `polygon_points`; both call sites can use `to_msg`.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/traffic_light_compliance.cpp
```diff
@@ -41,6 +42,57 @@ namespace
...
+{
+  geometry_msgs::msg::Point msg;
+  msg.x = point.x();
+  msg.y = point.y();
+  msg.z = z;
+  return msg;
+}
+
+std::vector<geometry_msgs::msg::Point> polygon_points(
```
go-sakayori: `metric_utils.cpp::polygon_to_points(Polygon2d, z)` is the same conversion (added in this PR). Including `metrics/geometry/metric_utils.hpp` and calling it directly would drop ~12 lines of duplication.

planning/autoware_planning_data_analyzer/src/metrics/trajectory_metrics.cpp
```diff
@@ -49,6 +50,54 @@ using autoware::route_handler::RouteHandler;
...
 
+geometry_msgs::msg::Point to_msg_point(
+  const geometry_msgs::msg::Point & point, const double z_offset = 0.0)
+{
+  geometry_msgs::msg::Point msg = point;
+  msg.z += z_offset;
+  return msg;
+}
+
+geometry_msgs::msg::Point to_msg_point(
```
go-sakayori: Anonymous-namespace `to_msg_point(Point2d, z)` here is identical to the one added in `metric_utils.cpp:93`. Either include `metrics/geometry/metric_utils.hpp` and reuse the public version, or inline the underlying `autoware_utils_geometry::to_msg(point.to_3d(z))` call — same idea applies to the three small helpers below (`lanelet_polygon_points`, `polygon_points`, `centerline_points`), which only differ in the lanelet container being iterated.

planning/autoware_planning_data_analyzer/src/metrics/trajectory_metrics.cpp
```diff
@@ -49,6 +50,54 @@ using autoware::route_handler::RouteHandler;
...
+  const autoware_utils_geometry::Point2d & point, const double z = 0.0)
+{
+  geometry_msgs::msg::Point msg;
+  msg.x = point.x();
+  msg.y = point.y();
+  msg.z = z;
+  return msg;
+}
+
+std::vector<geometry_msgs::msg::Point> lanelet_polygon_points(
```
go-sakayori, line 73: `lanelet_polygon_points` / `polygon_points` / `centerline_points` all (a) discard the lanelet point's real `z` and (b) reconstruct a `Point2d` only to push it back through the local `to_msg_point`. `autoware_utils_geometry::to_msg(Point3d{point.x(), point.y(), z})` (or `Point3d{x, y, point.z()}` if you want to preserve elevation) does the same in one call and removes the three near-identical loops.

go-sakayori (changes requested):

planning/autoware_planning_data_analyzer/src/metrics/epdms/debug/epdms_debug_writer.cpp
bskkimm: Addressed by reworking the helper to reuse `create_deleted_default_marker` as the base marker while preserving the existing `DELETEALL` action explicitly. This keeps the debug clear behavior unchanged across dependency versions.

planning/autoware_planning_data_analyzer/src/metrics/epdms/debug/epdms_debug_writer.cpp
bskkimm: Addressed. The local Polygon2d conversion was removed and the debug writer now reuses `metric_utils::polygon_to_points`.

planning/autoware_planning_data_analyzer/src/metrics/geometry/metric_utils.cpp
bskkimm: Addressed. `object_id_to_string` now uses `autoware_utils_uuid::to_hex_string` while preserving the existing `invalid` sentinel.

planning/autoware_planning_data_analyzer/src/metrics/geometry/metric_utils.cpp
bskkimm: Addressed for the duplicate call sites touched here. TLC now inlines `autoware_utils_geometry::to_msg(Point3d{...})`, and trajectory lanelet conversions were consolidated into one helper.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/no_at_fault_collision.cpp
bskkimm: Addressed. The manual yaw/sin/cos transform was replaced with `autoware_utils_geometry::calc_offset_pose` for both front bumper points.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/traffic_light_compliance.cpp
bskkimm: Addressed. The TLC point conversion now uses `autoware_utils_geometry::to_msg(Point3d{...})` directly.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/traffic_light_compliance.cpp
bskkimm: Addressed. The TLC local polygon conversion was removed and replaced with `metric_utils::polygon_to_points`.

planning/autoware_planning_data_analyzer/src/metrics/trajectory_metrics.cpp
bskkimm: Addressed. The duplicated Point2d conversion was removed; the lanelet/debug conversions now use a single local range helper with `autoware_utils_geometry::to_msg(Point3d{...})`.

planning/autoware_planning_data_analyzer/src/metrics/trajectory_metrics.cpp
bskkimm: Addressed. The three near-identical loops were consolidated into one templated `lanelet_points_to_msg` helper, preserving the previous behavior of using the supplied debug z value.

planning/autoware_planning_data_analyzer/src/metrics/trajectory_metrics.cpp
```diff
@@ -352,6 +393,46 @@ TrajectoryPointMetrics calculate_trajectory_point_metrics(
...
+      ddc_result.worst_window_end_time_s;
+    metrics.driving_direction_compliance_debug.worst_window_sample_count =
+      ddc_result.worst_window_sample_count;
+    metrics.driving_direction_compliance_debug.window_progress_m =
+      ddc_result.max_oncoming_progress_m;
+    for (size_t i = 0; i < driving_direction_evaluation_points.size(); ++i) {
+      const auto & evaluation_point = driving_direction_evaluation_points.at(i);
+      const auto & trajectory_point = trajectory.points.at(i);
+      const auto local_context =
+        compute_driving_direction_local_context(trajectory_point.pose, route_handler)
```
go-sakayori: This per-point `compute_driving_direction_local_context(...)` loop (and the `route_lane_polygons` / `intersection_lane_polygons` it materializes from `polygon3d()`) is not gated by `debug_topics_enabled` — `calculate_trajectory_point_metrics` has no knowledge of the flag, so this expensive lanelet querying runs on every evaluation even when no debug topics are written. It also looks like it recomputes context the subscore already derived from the pre-built `driving_direction_evaluation_points`.

Consider threading the enabled flag (or per-subscore enable bits) into `calculate_trajectory_point_metrics` and guarding this block. Same applies to the NC/TTC/LK debug accumulation. Could you re-measure the 197s-vs-212s numbers with the collection actually gated off?

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/no_at_fault_collision.cpp
```diff
@@ -218,17 +372,27 @@ NoAtFaultCollisionResult calculate_no_at_fault_collision(
...
         }
+        result.debug_info.events.push_back(debug_event);
         continue;
       }
+
+      result.debug_info.events.push_back(debug_event);
     }
   }
 
+  fill_horizon_debug_footprints(
```
go-sakayori: `fill_horizon_debug_footprints` iterates every trajectory point × every engaged object, re-interpolating object states and running `bg::intersects` + `overlap_polygons_to_points`, and it runs unconditionally as part of scoring (it only early-returns when `events` is empty). When `debug_topics_enabled` is false this work is wasted. Suggest gating it behind the debug flag.

planning/autoware_planning_data_analyzer/src/metrics/epdms/debug/epdms_debug_writer.cpp
```diff
@@ -0,0 +1,1069 @@
...
+  const TrajectoryPointMetrics & metrics, const rclcpp::Time & timestamp)
+{
+  return nlohmann::json{
+    {"trajectory_stamp_sec", timestamp.seconds()},
+    {"score", metrics.ego_progress},
+    {"available", metrics.ego_progress_available},
+    {"reason", metrics.ego_progress_reason},
+    {"raw_progress_m", metrics.ego_progress_raw_m},
+    {"best_raw_progress_m", metrics.ego_progress_best_raw_m},
+    {"masked_progress_m", metrics.ego_progress_denominator_m},
```
go-sakayori: `masked_progress_m` and `denominator_m` carry the same value (`ego_progress_denominator_m`), and just below `multiplicative_mask` / `mask` are both `ego_progress_mask`. Worth dropping the redundant pair. Also `{"threshold_m", 5.0}` at line 449 is a hardcoded magic number — if it mirrors an EP parameter, reference that instead of duplicating the literal.

go-sakayori (commented):
A few inline notes. Overall the feature is cleanly factored and correctly defaults off — the main thing worth addressing before merge is that the debug *data collection* (not just the bag writing) appears to run unconditionally, so the cost is paid even when `debug_topics_enabled` is false.

planning/autoware_planning_data_analyzer/src/metrics/trajectory_metrics.cpp
bskkimm: Addressed in the latest push. The DDC local-context/polygon debug collection is now gated by `debug_topics_enabled && driving_direction_compliance`, while the score path still computes the required evaluation context.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/no_at_fault_collision.cpp
bskkimm: Addressed in the latest push. `fill_horizon_debug_footprints` is now controlled by a `collect_debug` flag, so the trajectory/object footprint and overlap pass is skipped when debug topics are disabled.

planning/autoware_planning_data_analyzer/src/metrics/epdms/debug/epdms_debug_writer.cpp
bskkimm: Addressed in the latest push. Removed the duplicate EP summary fields and replaced the hardcoded `5.0` threshold with the shared EP threshold constant.

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
```diff
@@ -530,7 +476,8 @@ void OpenLoopEvaluator::evaluate(
             : std::vector<TimedTrackedObjects>{};
         auto trajectory_metrics = metrics::calculate_trajectory_point_metrics(
           eval_data.synchronized_data, route_handler_, history_comfort_params_,
-          lane_keeping_params_, driving_direction_params_, vehicle_info_, future_objects);
+          lane_keeping_params_, driving_direction_params_, vehicle_info_, future_objects,
+          debug_topics_enabled_ && enabled_metrics_.driving_direction_compliance);
```
go-sakayori: The gating fixes are correct in isolation, but this single `collect_debug` argument now drives **two independent subscores** while being derived only from the DDC enable bit. Inside `calculate_trajectory_point_metrics` the same flag is forwarded to `calculate_no_at_fault_collision(..., collect_debug)` (gating `fill_horizon_debug_footprints`) *and* used to guard the DDC per-point loop.

Consequence: with `debug_topics_enabled` on, NC enabled, but **DDC disabled**, `collect_debug` is `false`, so NC horizon footprints are never collected — and `write_epdms_point_debug_topics_to_bag` still emits `/debug/epdms/nc/ego_footprints`, `object_footprints`, and `overlap_areas` as DELETEALL-only (empty) arrays. The NC collision_summary still works (events are collected unconditionally), so the breakage is silent.

Suggest deriving the flags per subscore instead of sharing one, e.g. pass `debug_topics_enabled_ && enabled_metrics_.no_at_fault_collision` to the NC call and `debug_topics_enabled_ && enabled_metrics_.driving_direction_compliance` to the DDC block — most cleanly by replacing the single `collect_debug` bool with a small struct (or reusing `EpdmsDebugEnabledMetrics`) so each subscore reads its own bit.

go-sakayori (commented):
Thanks for the quick turnaround — all three of my earlier notes are addressed, and sharing `kEgoProgressDistanceThresholdM` between `ego_progress` and the debug writer is a nicer fix than I suggested. One new issue introduced by the gating change (inline), plus a minor remaining note.

Minor (non-blocking): LK debug collection in `calculate_lane_keeping_result` still runs unconditionally — it reserves/pushes a `LaneKeepingDebugSample` per point and copies `reference_centerline` into each. Much cheaper than the DDC/NC paths you just gated, so fine to leave, but it could take the same `collect_debug` treatment later.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/lane_keeping.cpp
```diff
@@ -22,32 +22,39 @@
...
 
   std::optional<rclcpp::Duration> violation_start_time;
+  std::optional<std::size_t> violation_start_index;
+  double max_violation_duration = 0.0;
+  double peak_abs_lateral_deviation = 0.0;
+  bool failure_recorded = false;
   std::optional<double> queue_release_until_s;
 
-  const auto reset_violation_run = [&]() { violation_start_time.reset(); };
+  result.debug.samples.reserve(evaluation_points.size());
```
go-sakayori: Now that NC and DDC debug collection are gated behind `collect_debug`, this is the remaining subscore that still builds its debug payload unconditionally: it reserves and pushes a `LaneKeepingDebugSample` for every evaluation point, copying `reference_centerline` into each sample, even when debug topics are disabled. Cheaper than the DDC/NC paths you just gated, so non-blocking — but for consistency it would be worth threading the same `collect_debug` flag through `calculate_lane_keeping_result` and skipping the per-sample accumulation (the `score` computation doesn't need it).

bskkimm: Addressed in the latest push as well. LK now threads its own debug gate into calculate_lane_keeping_result and skips per-sample debug accumulation when LK debug topics are disabled, while leaving the score computation unchanged.

planning/autoware_planning_data_analyzer/src/open_loop_evaluator.cpp
bskkimm: Addressed in the latest push. I split the shared debug gate into per-subscore flags by passing a small trajectory-metrics debug struct, so NC debug collection now depends on the NC enable bit and DDC debug collection depends on the DDC enable bit independently.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/traffic_light_compliance.cpp
```diff
@@ -248,12 +292,17 @@ TrafficLightComplianceResult calculate_traffic_light_compliance(
...
   const bool can_reuse_evaluations =
     evaluations != nullptr && evaluations->size() == trajectory.points.size();
+  bool violation_found = false;
 
   for (std::size_t point_index = 0; point_index < trajectory.points.size(); ++point_index) {
     const auto & point = trajectory.points.at(point_index);
     const auto ego_polygon = can_reuse_evaluations
                                ? evaluations->at(point_index).ego_polygon
                                : create_pose_footprint(point.pose, local_footprint);
+    result.debug_info.ego_horizon_footprints.push_back(
```
go-sakayori: TLC has no `collect_debug` gate, so this line allocates on every evaluation even with debug off.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/no_at_fault_collision.cpp
```diff
@@ -196,20 +341,30 @@ NoAtFaultCollisionResult calculate_no_at_fault_collision(
         continue;
       }
 
-      const auto collision_type = classify_collision(point, *object_state, vehicle_info);
-      const bool front_or_stopped_track = collision_type == CollisionType::ActiveFront ||
-                                          collision_type == CollisionType::StoppedTrack;
-      const bool lateral_collision = collision_type == CollisionType::ActiveLateral;
+      const auto collision = classify_collision(point, *object_state, vehicle_info);
+      auto debug_event =
+        make_debug_event(query_time_s, point, ego_polygon, *object_state, collision);
```
go-sakayori: This is run unconditionally. Can we gate this too?

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ttc_within_bound.cpp
```diff
@@ -206,6 +326,15 @@ TTCWithinBoundResult calculate_ttc_within_bound(
         const bool ahead = is_agent_ahead_nuplan(projected_pose, object_state->pose);
         const bool behind = is_agent_behind_nuplan(projected_pose, object_state->pose);
         if (ahead || (bad_or_intersection && !behind)) {
+          auto debug_event = make_debug_event(
+            time_s, future_offset_s, query_time_s, bad_or_intersection, multiple_lanes,
+            non_drivable_area, ego_in_intersection, ahead, behind, projected_pose, ego_polygon,
+            *object_state);
+          result.debug_info.events.push_back(debug_event);
+          populate_ttc_debug_horizon(
```
go-sakayori: No debug flag here.

go-sakayori (commented):
Some subscore build debug data unconditionally.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/ttc_within_bound.cpp
bskkimm: Addressed in b44f7761. TTC now receives the per-subscore collect_debug gate and only builds debug events, horizon footprints, and overlap payloads when debug collection is enabled. Failure detection and score updates remain unchanged.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/no_at_fault_collision.cpp
bskkimm: Addressed in b44f7761. NC now creates and stores debug events/horizon payloads only when collect_debug is true. Collision classification, at-fault recording, and score updates still run unconditionally.

planning/autoware_planning_data_analyzer/src/metrics/epdms/subscores/traffic_light_compliance.cpp
bskkimm: Addressed in b44f7761. TLC now receives the per-subscore collect_debug gate and only populates TLC debug payloads, stop lines, ego horizon footprints, and failure anchors when debug collection is enabled. Score and availability logic remain unchanged.

go-sakayori (approved):

PR #455, opened 2026-06-26
ci(check-build-depends): use universe-dependencies container

mitsudome-r:
## Description
- Update the `check-build-depends` CI workflow to build against ROS 2 Jazzy instead of Humble, matching the project's current target distribution.
- Replace the plain `ros:humble` container with the prebuilt `ghcr.io/autowarefoundation/autoware:universe-dependencies-jazzy` image, so Autoware's universe dependencies are already present during the build-depends check.
- Bump the runner to `ubuntu-24.04` and simplify the matrix by inlining `build-depends-repos: build_depends.repos` (the per-distro override is no longer needed with a single distro).

## How was this PR tested?
- [ ] Confirm the `check-build-depends` workflow runs on this PR and resolves the `universe-dependencies-jazzy` container.
- [ ] Verify the build-depends check passes against `build_depends.repos` under Jazzy.

## Notes for reviewers

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/workflows/check-build-depends.yaml
+++ b/.github/workflows/check-build-depends.yaml
@@ -7,17 +7,16 @@ on:
 
 jobs:
   check-build-depends:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     container: ${{ matrix.container }}
     strategy:
       fail-fast: false
       matrix:
         rosdistro:
-          - humble
+          - jazzy
         include:
-          - rosdistro: humble
-            container: ros:humble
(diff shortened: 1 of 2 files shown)


mitsudome-r:
I cherry picked https://github.com/autowarefoundation/autoware_tools/pull/449/changes to trigger the workflow. 
I will revert it once I confirm that the CI passes.
https://github.com/autowarefoundation/autoware_tools/actions/runs/28239748932/job/83663360871?pr=455

PR #456, opened 2026-06-26
ci: use underlay workspace

mitsudome-r:
## Description
This is part of https://github.com/orgs/autowarefoundation/discussions/7096

Switches CI to build/test on top of the pre-built `universe-dependencies` Docker image and its `/opt/autoware` underlay workspace, instead of rebuilding core packages from source.

- Pass `underlay-workspace: /opt/autoware` to the build/test actions in `build-and-test.yaml`, `build-and-test-differential.yaml`, and `check-build-depends.yaml` so colcon builds against the pre-built core packages in the image.
- Migrate `check-build-depends` from Humble to Jazzy: `ubuntu-24.04` runner and the `ghcr.io/autowarefoundation/autoware:universe-dependencies-jazzy` container, dropping the per-matrix `build-depends-repos` in favor of a fixed `build_depends.repos`.
- Remove all `core/*` (and `autoware_cmake`) entries from `build_depends.repos`, since the core packages are now provided pre-built by the CI image and pinned by the image tag. Added a comment documenting that core repos must not be re-added here to avoid redundant rebuilds.

## How was this PR tested?

- [x] `build-and-test` workflows pass with the Jazzy `universe-dependencies` image and `/opt/autoware` underlay. https://github.com/mitsudome-r/autoware_tools/actions/runs/28241030822
- [x] `check-build-depends` workflow passes on `ubuntu-24.04` / `jazzy` against `build_depends.repos`. https://github.com/autowarefoundation/autoware_tools/actions/runs/28240888552/job/83667191031?pr=456

## Notes for reviewers

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/workflows/build-and-test-differential.yaml
+++ b/.github/workflows/build-and-test-differential.yaml
@@ -55,6 +55,21 @@ jobs:
         id: get-modified-packages
         uses: autowarefoundation/autoware-github-actions/get-modified-packages@v1
 
+      - name: Remove autoware_core packages from underlay
+        run: |
+          tmp_dir=$(mktemp -d)
+          git clone --depth 1 --branch main \
+            https://github.com/autowarefoundation/autoware_core.git "$tmp_dir"
+          for pkg_xml in $(find "$tmp_dir" -name package.xml); do
+            pkg_name=$(xmllint --xpath 'string(//package/name)' "$pkg_xml")
+            if [ -d "/opt/autoware/$pkg_name" ]; then
+              echo "Removing /opt/autoware/$pkg_name"
+              sudo rm -rf "/opt/autoware/$pkg_name"
+            fi
+          done
+          rm -rf "$tmp_dir"
+        shell: bash
+
       - name: Build
         if: ${{ steps.get-modified-packages.outputs.modified-packages != '' }}
         uses: autowarefoundation/autoware-github-actions/colcon-build@v1
--- a/.github/workflows/build-and-test-differential.yaml
+++ b/.github/workflows/build-and-test-differential.yaml
@@ -62,6 +62,7 @@ jobs:
           rosdistro: ${{ matrix.rosdistro }}
           target-packages: ${{ steps.get-modified-packages.outputs.modified-packages }}
           build-depends-repos: build_depends.repos
+          underlay-workspace: /opt/autoware
 
       - name: Test
         id: test
@@ -71,7 +72,7 @@ jobs:
           rosdistro: ${{ matrix.rosdistro }}
           target-packages: ${{ steps.get-modified-packages.outputs.modified-packages }}
           build-depends-repos: build_depends.repos
-
+          underlay-workspace: /opt/autoware
       - name: Upload coverage to CodeCov
         if: ${{ steps.test.outputs.coverage-report-files != '' }}
         uses: codecov/codecov-action@v3
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -47,7 +47,7 @@ jobs:
           rosdistro: ${{ matrix.rosdistro }}
           target-packages: ${{ steps.get-self-packages.outputs.self-packages }}
           build-depends-repos: ${{ matrix.build-depends-repos }}
-
+          underlay-workspace: /opt/autoware
       - name: Test
         if: ${{ steps.get-self-packages.outputs.self-packages != '' }}
         id: test
@@ -56,7 +56,7 @@ jobs:
           rosdistro: ${{ matrix.rosdistro }}
           target-packages: ${{ steps.get-self-packages.outputs.self-packages }}
           build-depends-repos: ${{ matrix.build-depends-repos }}
-
+          underlay-workspace: /opt/autoware
       - name: Upload coverage to CodeCov
         if: ${{ steps.test.outputs.coverage-report-files != '' }}
         uses: codecov/codecov-action@v3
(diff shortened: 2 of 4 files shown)


.github/workflows/check-build-depends.yaml
```diff
@@ -34,4 +33,5 @@ jobs:
         with:
           rosdistro: ${{ matrix.rosdistro }}
           target-packages: ${{ steps.get-self-packages.outputs.self-packages }}
-          build-depends-repos: ${{ matrix.build-depends-repos }}
+          build-depends-repos: build_depends.repos
+          underlay-workspace: /opt/autoware
```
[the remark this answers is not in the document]
mitsudome-r: Fixed in https://github.com/autowarefoundation/autoware_tools/pull/456/commits/19554ab397adbedf305b66034966181dc24050f3

.github/workflows/build-and-test-differential.yaml
xmfcx, line 57: Keeping `core/autoware_core` at `main` in `build_depends.repos` makes sense to me.

The one overlap to handle is `autoware_core`: the `universe-dependencies` image ships it pre-built at the image's tagged version in `/opt/autoware`, and you now also rebuild it from `main` in the overlay. The overlay should win, but stale or renamed packages left in the tagged underlay can leak through as ghost packages or duplicate-package warnings.

I'd add a step just before the build that strips the `autoware_core` packages out of `/opt/autoware`, the same idea as autoware_core's own CI ([ref](https://github.com/autowarefoundation/autoware_core/blob/97ebb6c0371b065ed984049ca1c1d62999519447/.github/workflows/build-and-test-reusable.yaml#L121-L131)). Since `autoware_core` isn't in this checkout, the step shallow-clones it first to learn the package names:

```suggestion

      - name: Remove autoware_core packages from underlay
        run: |
          tmp_dir=$(mktemp -d)
          git clone --depth 1 --branch main \
            https://github.com/autowarefoundation/autoware_core.git "$tmp_dir"
          for pkg_xml in $(find "$tmp_dir" -name package.xml); do
            pkg_name=$(xmllint --xpath 'string(//package/name)' "$pkg_xml")
            if [ -d "/opt/autoware/$pkg_name" ]; then
              echo "Removing /opt/autoware/$pkg_name"
              sudo rm -rf "/opt/autoware/$pkg_name"
            fi
          done
          rm -rf "$tmp_dir"
        shell: bash

```

That way the `main` build is the only copy of `autoware_core` in scope, and the tagged underlay copies can't shadow it.

xmfcx (changes requested):

.github/workflows/build-and-test.yaml
xmfcx, line 42: Needed here too

mitsudome-r: removed from both build-and-test and check-build-depends in  https://github.com/autowarefoundation/autoware_tools/pull/456/commits/9f42ab6f12136861e41b49bac5dd7c4bc56f2b38

build_depends.repos
```diff
@@ -1,29 +1,7 @@
+# Do not add core (core/*) dependency repositories here.
+# Core packages are provided pre-built by the CI Docker image, and their versions
+# are defined by the image tag. Adding them here would cause redundant rebuilds.
```
xmfcx, line 3: Reword to resolve the contradiction with the `core/autoware_core: main` entry just below — it carves out autoware_core as the intentional exception:

```suggestion
# Core message/utility packages (autoware_msgs, autoware_utils, etc.) come pre-built
# from the CI Docker image, pinned by the image tag. Do not add them here; it would
# cause redundant rebuilds. autoware_core is the deliberate exception: it is rebuilt
# from main below, and the "Remove autoware_core packages from underlay" CI step drops
# the image's tagged copy so this source build is the only one in scope.
```

xmfcx (approved):
Thanks for working through this one, and for being patient with the back-and-forth on the underlay shadowing. The result is clean.

Approving. The underlay switch itself was already green on humble + jazzy in the full `build-and-test` run, and [this log](https://github.com/autowarefoundation/autoware_tools/actions/runs/28250453483/job/83700371994#step:9:18) shows the new step correctly stripping the tagged `autoware_core` copies out of `/opt/autoware` (`autoware_core`, `autoware_motion_utils`, `autoware_ekf_localizer`, …) before the build. That's enough to be confident the `main` overlay won't be shadowed by the underlay, so the duplication concern from my earlier review is resolved. Thanks for applying it to all three workflows.

A few non-blocking follow-ups (fine to defer or skip):

- **`build_depends.repos` header vs content.** The comment says "Do not add core (core/*) … here," but `core/autoware_core: main` stays right below it, which reads as a contradiction (Copilot tripped on it too). Left an inline suggestion to reword the header so it carves out `autoware_core`.
- **The remove step is unconditional** while Build/Test are gated on `*-packages != ''`. On a PR like this one (no packages modified) it still clones `autoware_core` and `rm -rf`s the underlay, then skips the build. Gating it with the same `if:` would save the wasted work.
- **It's now copy-pasted in three workflows** with a hardcoded URL/branch. A local composite action (like `.github/actions/tune-dds-network`) would keep them from drifting. Could be a separate cleanup PR.
- **Heads-up, not an ask:** `core/autoware_core: main` now compiles against the msgs/utils baked into the image tag, so a future core `main` that needs newer msgs than the image carries will break the build until the image is rebumped. Inherent to the approach, just worth being aware of.

Nits (purely cosmetic, take them or leave them):
- `check-build-depends.yaml` uses `actions/checkout@v3` while the other two are on `@v6`.
- A couple of blank lines between steps got dropped in the build workflows.

PR #457, opened 2026-06-30
feat(planning_data_analyzer): add EPDMS calculation gate

bskkimm:
## Description

Adds a DLR-controllable gate for EPDMS work in `autoware_planning_data_analyzer`.

The new `open_loop.enable_epdms_calculation` parameter and `enable_epdms_calculation` launch argument default to `true` to preserve existing behavior. When disabled, the open-loop evaluator skips EPDMS subscore calculation, human-reference/filter aggregation, synthetic EPDMS aggregation, EPDMS metric topics, EPDMS trajectory metric topics, and `/debug/epdms/*` topic registration/output while preserving base trajectory metrics.

This is intended for DLR post-processing where optimized and raw trajectories are analyzed separately but EPDMS should only be calculated once.

## Changes

- Add `open_loop.enable_epdms_calculation` config parameter.
- Add `enable_epdms_calculation` launch argument.
- Wire the parameter through the analyzer node to `OpenLoopEvaluator`.
- Skip EPDMS calculation/output/debug paths when disabled.
- Add a unit test for result-topic gating.

## Testing

- `git diff --check` passed.
- `clang-format` applied to touched C++ files.
- `colcon build --packages-select autoware_planning_data_analyzer` was attempted locally, but this checkout is blocked by the local underlay lanelet API mismatch before reaching this change:
  - `get_arc_coordinates_on_ego_centerline` is not available in the locally sourced `autoware::experimental::lanelet2_utils`.

Please run the standard upstream CI, including the differential build/test check used on PR #452 (`tag:run-build-and-test-differential`).

### DIFF
--- a/planning/autoware_planning_data_analyzer/README.md
+++ b/planning/autoware_planning_data_analyzer/README.md
@@ -82,6 +82,7 @@ Key parameters in `config/planning_data_analyzer.param.yaml`:
 - `evaluation_interval_ms`: Sampling interval (default: 100ms)
 - `sync_tolerance_ms`: Time synchronization tolerance (default: 100ms)
 - `trajectory_topic`: Trajectory topic to evaluate
+- `open_loop.enable_epdms_calculation`: Enables EPDMS subscore, synthetic EPDMS, and EPDMS debug calculation
 - `open_loop.*`: Open-loop specific settings
 
 ---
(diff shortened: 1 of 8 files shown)


bskkimm:
Closing temporarily to verify local build behavior with a temporary lanelet API compatibility check before reopening a clean PR.

PR #466, opened 2026-09-16
fix(deviation_evaluator, parameter_estimator): migrate map launch to component_map.launch.xml

hakuturu583:
## Description

Follow-up to #465. That PR renamed the `tier4_map_launch` dependency to `autoware_map_launch`, but the change did not stick: the pre-commit **check package depends** hook re-added `tier4_map_launch` (see the autofix commit on #465), because the launch files here still include `$(find-pkg-share tier4_map_launch)/launch/map.launch.xml`. So `main` ended up with both `autoware_map_launch` and the removed-from-upstream `tier4_map_launch`, and rosdep still cannot resolve `tier4_map_launch` — keeping autoware.universe's `build-and-test-packages-above-differential` red.

[autowarefoundation/autoware_launch#1980](https://github.com/autowarefoundation/autoware_launch/pull/1980) renamed `tier4_map_launch` → `autoware_map_launch` **and decomposed** its `map.launch.xml` aggregator into per-node launch files (`map_loader/*`, `map_projection_loader/*`, `map_tf_generator/*`, `map_hash_generator/*`). There is no more single `map.launch.xml`; the exported map composition now lives at `autoware_launch/launch/components/component_map.launch.xml`.

## Change

Switch both tools' map include from `tier4_map_launch/launch/map.launch.xml` to **`autoware_launch/launch/components/component_map.launch.xml`**:

- `localization/deviation_estimation_tools/deviation_evaluator/launch/deviation_evaluator.launch.xml`
- `vehicle/parameter_estimator/launch/parameter_estimator_with_simulation.launch.xml`

`component_map.launch.xml` resolves the individual map files from `map_path` and loads the map parameters internally from `autoware_map_config`, so the long list of `*_param_path` args is no longer needed — only `map_path`, `lanelet2_map_file` and `pointcloud_map_file` are passed (all already declared in these launch files).

With the `tier4_map_launch` / `autoware_map_launch` references gone from the launch files, both are dropped from `package.xml`. `autoware_launch` is already an `exec_depend`, and it pulls in the map launchers transitively, so the map nodes stay available at runtime and the dependency hook has nothing to re-add.


sasakisasaki (approved):
Reviewed, and also I verified that the following command succeeded based on this PR fix 👍 

```bash
$ git clone git@github.com:autowarefoundation/autoware.git
$ cd autoware
$ vcs import src/ < repositories/autoware.repos && vcs import src/ < repositories/autoware-nightly.repos
$ vcs import src/ < repositories/simulator.repos
$ vcs import src/ < repositories/simulator-nightly.repos
$ vcs import src/ < repositories/tools.repos
$ vcs import src/ < repositories/tools-nightly.repos
$ rosdep install -y --from-paths ./ --ignore-src --rosdistro $ROS_DISTRO
```

mitsudome-r (approved):

